PR #6, opened 2026-03-12
fix(chat): prevent textarea scroll content under bottom toolbar

Jasion-han:
Fix MessageInput textarea so long text never scrolls under the bottom toolbar (attach/mode + send/cancel) while keeping the original auto-grow behavior.


xintaofei:
佬，我这里滚动条下半截没了：
image

PR #55, opened 2026-04-02
refactor(chat): extract and reuse permission action buttons

isYangs:
## Summary

Refactor the chat permission dialog by extracting permission action buttons into a dedicated component and localizing the action labels.

## Changes

- add a reusable `PermissionActions` component for permission responses
- move permission action button rendering out of `PermissionDialog`
- localize permission action labels across supported languages
- keep the existing permission response flow unchanged while simplifying the dialog component

### DIFF
--- a/src/components/chat/permission-actions.tsx
+++ b/src/components/chat/permission-actions.tsx
@@ -1,0 +1,80 @@
+"use client"
+
+import { useTranslations } from "next-intl"
+import { Button } from "@/components/ui/button"
+import type { PermissionOptionInfo } from "@/lib/types"
+
+type PermissionActionVariant = "default" | "outline"
+
+interface PermissionActionsProps {
+  options: PermissionOptionInfo[]
+  onRespond: (optionId: string) => void
+}
+
+const KIND_LABEL_KEYS: Record<string, string> = {
+  allow_once: "allowOnce",
+  allow_always: "allowAlways",
+  reject_once: "rejectOnce",
+  reject_always: "rejectAlways",
+}
+
+const KIND_VARIANTS: Record<string, PermissionActionVariant> = {
+  allow_once: "default",
+  allow_always: "default",
+  reject_once: "outline",
+  reject_always: "outline",
+}
+
+function extractDetail(name: string): string | undefined {
+  const match = name.match(/`([^`]+)`/)
+  return match?.[1]
+}
+
+export function PermissionActions({
+  options,
+  onRespond,
+}: PermissionActionsProps) {
+  const t = useTranslations("Folder.chat.permissionDialog.actions")
+
+  return (
+    <div className="mt-3 flex flex-wrap gap-2">
+      {options.map((opt) => {
+        const variant: PermissionActionVariant =
+          KIND_VARIANTS[opt.kind] ??
+          (opt.kind.startsWith("reject") ? "outline" : "default")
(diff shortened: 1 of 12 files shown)


JiwaniZakir (commented):
The refactored `PermissionActions` component introduces a subtle regression: the fallback button (no `detail`) uses a fixed `className="h-9"`, but the original code in `permission-dialog.tsx` used `h-auto min-h-9 whitespace-normal break-words` to accommodate longer translated labels. With 10+ locales added in this PR, some translations (e.g., Arabic `رفض دائمًا`) may overflow or get clipped at a fixed `h-9` — the `max-w-full basis-full` treatment is only applied when `detail` is present.

The `t(labelKey as never)` cast in `permission-actions.tsx` suppresses TypeScript's key-safety checks on the translation function. Consider defining a proper union type from `KIND_LABEL_KEYS` values (or a typed `const` assertion) so the compiler can verify those keys exist in the message catalog — especially important since this component now shares translation keys across all locales added in this diff.

Finally, `extractDetail` only runs when `labelKey` is defined, meaning any unrecognized `opt.kind` values whose names contain backtick-delimited content will silently render the raw `opt.name` string without the split label/detail layout. Whether that's intentional should be documented with a comment, since it's a non-obvious branch in the rendering logic.

isYangs:
> 重构后的`PermissionActions`组件引入了一个细微的回归问题：备用按钮（无`detail`）使用了固定长度`className="h-9"`，但原始代码`permission-dialog.tsx`用于容纳更长的翻译标签。由于此 PR 中添加了 10 多个语言环境，某些翻译（例如阿拉伯语）可能会溢出或被截断到固定长度——这种处理方式仅在存在时才会生效。`h-auto min-h-9 whitespace-normal break-words``رفض دائمًا``h-9``max-w-full basis-full``detail`
> 
> 强制类型转换会抑制TypeScript 对翻译函数的键安全检查。请考虑从值定义一个合适的联合类型（或类型化断言），以便编译器可以验证这些键是否存在于消息目录中——这一点尤为重要，因为此组件现在会在所有新增语言环境中共享翻译键。`t(labelKey as never)``permission-actions.tsx``KIND_LABEL_KEYS``const`
> 
> 最后，该`extractDetail`函数仅`labelKey`在定义了 `<value>` 时运行，这意味着任何`opt.kind`名称包含反引号分隔内容的未识别值都将静默渲染原始`opt.name`字符串，而不会使用拆分的标签/详情布局。这是否是预期行为应该在注释中加以说明，因为这是渲染逻辑中一个不太明显的分支。

Thanks for the thorough review! All three points are valid and have been addressed:

**Fixed height regression** — Both button variants now use `h-auto min-h-9` instead of the fixed `h-9`. The fallback button also restores `whitespace-normal break-words` to match the original behavior, so longer translated labels (e.g., Arabic) can wrap naturally.
**Type safety** — `KIND_LABEL_KEYS` is now declared with `as const`, and an `ActionLabelKey` union type is extracted from its values. The `as never` cast has been replaced with `as ActionLabelKey`, giving the compiler a narrower constraint to work with.
**Fallback behavior for unrecognized kinds** — This is intentional: when `opt.kind` isn't in our known set, we have no translation key, so rendering the raw `opt.name` without splitting is the correct degradation path. Added a comment to make this explicit.

PR #111, opened 2026-04-25
fix(acp): dedupe slash command results

MoozLee:
## Summary
- Deduplicate ACP slash command records by command name before UI consumers render them.
- Preserve existing fuzzy search behavior and first-occurrence ordering.

## Test plan
- pnpm eslint .
- pnpm build
image
image


xintaofei:
感谢pr

PR #272, opened 2026-06-11
Add pagination for conversation history loading

ijry:
## Summary
- add backend pagination params and metadata to get_folder_conversation
- load only the newest page by default, with incremental history loading in the conversation runtime
- add a UI action to load earlier messages for conversation detail, with i18n strings and updated tests

## Testing
- pnpm exec tsc --noEmit
- pnpm exec vitest run src/contexts/conversation-runtime-context.test.tsx src/components/conversations/conversation-detail-panel-layout.test.ts src/components/message/sub-agent-session-dialog.test.tsx


xintaofei:
Thank you for your pull request. The current virtual list performs very well and does not require pagination. In fact, pagination would lose some functionality, such as message navigation and conversation location.

PR #288, opened 2026-06-17
fix(chat): show full model name on hover

H1d3rOne:
## Summary
- Add hover title text for model options in the chat input model selector.
- Add hover title text for the selected model label when it is truncated.

## Why
When choosing a model, long model names are truncated with ellipsis, making it hard to tell which model is which. Hovering now shows the complete model name.

## Notes
- No port-related configuration files were changed.

## Checks
- pnpm eslint src/components/chat/session-config-selector.tsx src/components/chat/dropdown-radio-item-content.tsx
- git diff --check


xintaofei:
感谢pr。

PR #308, opened 2026-06-28
Expand local Codex session scanning

hight456789:
Adds multi-root Codex history scanning for active and archived sessions while preserving Codeg's existing UI.

Reverse sync behavior: imported Codex/Claude conversations continue through native resume flows by passing external_id as the native session id to Codeg's ACP connect path. Native tools remain the transcript writers. Codeg metadata stays in Codeg storage keyed by (agent_type, external_id).

Verification:
- PASS: node node_modules\\vitest\\vitest.mjs run src\\components\\conversations\\conversation-detail-panel-layout.test.ts
- PASS: node node_modules\\vitest\\vitest.mjs run src\\contexts\\acp-connections-context.test.tsx
- NOT RUN: cargo test codex (cargo is not installed or not on PATH in this environment)


hight456789:
Closing for now; this branch will be used internally first.

PR #368, opened 2026-07-22
feat(chat): fuzzy subsequence match for slash commands

asxuen:
## Summary
- Replace substring-only `/` command (and `$` skill) filtering with ranked fuzzy matching: exact → prefix → substring → subsequence.
- Lets inputs like `/bmadhelp` and `/bmhp` match hyphenated command names such as `bmad-help`.
- Shared helper used by chat composer, `+` slash dropdown, and automation editor menus.

## Test plan
- [x] Unit tests: `pnpm exec vitest run src/lib/fuzzy-text-match.test.ts`
- [ ] In chat input, type `/bmadhelp` and `/bmhp` and confirm `bmad-help` appears
- [ ] Empty `/` still lists all commands in original order
- [ ] Exact/prefix hits rank above pure subsequence matches
- [ ] `+` → Slash commands search and automation editor `/` behave the same

### DIFF
--- a/src/components/automations/composer-invocations.tsx
+++ b/src/components/automations/composer-invocations.tsx
@@ -16,6 +16,7 @@ import {
 } from "@/components/chat/composer/invocation-reference"
 import type { ReferenceAttrs } from "@/components/chat/composer/types"
 import { useAgentSkills } from "@/hooks/use-agent-skills"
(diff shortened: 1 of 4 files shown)


xintaofei:
感谢佬的PR

Issue #364, opened 2026-07-22
[Bug] ACP cancellation is finalized before agent confirmation, so Pi can keep running and late output is lost


Zker67:
## Summary

Codeg currently treats an ACP cancellation as completed immediately after enqueueing `session/cancel`, before the agent confirms cancellation through the original `session/prompt` response.

With Pi via `pi-acp`, this can produce a visible state split:

- Codeg marks the conversation `cancelled` and the UI stops showing the turn.
- The Pi process can continue running.
- Pi may later produce a complete final answer in its session file, but Codeg no longer surfaces it.

This issue is specifically about Codeg's cancellation confirmation and late-result reconciliation. The upstream model/API stream failure that initially caused the long stall is a separate problem.

## Environment

- Codeg: `0.21.4`
- OS: Windows 11
- Pi: `0.81.1`
- `pi-acp`: `0.0.31`
- Agent transport: ACP over stdio, Pi RPC child process

## Observed timeline

Local time UTC+8, July 22, 2026:

1. A long Pi turn encountered upstream Responses streaming failures, including an EOF and a truncated SSE JSON event.
2. At `14:38:30`, Codeg persisted the conversation as `cancelled`.
3. Pi continued working after that state transition:
   - `14:38:48`: Pi started another operation/compaction step.
   - `14:42:54`: Pi recorded another truncated SSE error.
   - `14:43:43`: Pi completed a normal final assistant response (`stopReason: stop`) and wrote it to its JSONL session.
4. The final response did not appear in Codeg because the Codeg-side turn had already been finalized as cancelled.
5. A process sample confirmed that the Pi RPC worker was responsive and mostly idle afterward; this was not a local CPU dead loop.

The Codeg database and Pi JSONL session therefore disagree about whether the turn was actually cancelled.

## Source analysis

At commit `f073b659e8732745fed42adae83d31219175b278`:

1. `ConnectionManager::cancel()` enqueues `ConnectionCommand::Cancel` and then eagerly changes the conversation row from `InProgress` to `Cancelled`:

https://github.com/xintaofei/codeg/blob/f073b659e8732745fed42adae83d31219175b278/src-tauri/src/acp/manager.rs#L1160-L1215

2. The active-turn cancel handler sends `CancelNotification`, but immediately emits a synthetic `TurnComplete { stop_reason: "cancelled" }` without waiting for the agent:

[...]

PR #370, opened 2026-07-22
fix(acp): bound Unix terminal cancellation

keh4l:
## Summary

- launch Unix ACP terminal commands in isolated process groups and clean them up with bounded `SIGINT` (500 ms), `SIGTERM` (1500 ms), then `SIGKILL` (2000 ms) escalation
- avoid holding the child mutex across exit waits, clean multiple session terminals concurrently, and cap Cancel's outer terminal-cleanup wait at 5 seconds
- always converge the connection to `TurnComplete(cancelled)` even when terminal cleanup times out, while reporting incomplete cleanup as a recoverable error
- track process-group ownership with leader/descendant/retired lease states, retire permanently on `ESRCH` or expiry, and serialize leader-exit observation with cleanup to prevent stale-PGID signalling

## Verification

- terminal-runtime tests — 29 passed, including a real stubborn process group, unrelated-session isolation, delayed release, lease expiry/`ESRCH`, and the leader-reap concurrency barrier
- Cancel convergence test — 1 passed
- `cargo check --no-default-features --bin codeg-server`
- `cargo test --no-default-features --bin codeg-server --lib` — 1650 passed, 1 ignored
- `cargo check --no-default-features --bin codeg-mcp`
- `cargo clippy --no-default-features --bin codeg-mcp -- -D warnings`
- no controlled-test child processes remained after the suite

## Scope and limitations

- Unix behavior is changed; Windows keeps the existing `kill_tree` path.
- This does **not** change agent acknowledgement or late-result reconciliation semantics tracked in #364. It addresses bounded terminal-process cleanup and connection-state convergence after Cancel.
- POSIX has no generation-bound process-group handle, so a final `probe`/`signal` syscall-level TOCTOU remains within the non-renewing five-second descendant lease. The previous unbounded stale-PGID window is removed.
- Desktop `cargo check` is not reported because this host lacks the GTK/GLib development packages required by `glib-sys`.
- Full-repository `cargo fmt --check` already fails on upstream `main`; this PR adds no new failing file and formats `terminal_runtime.rs`, which is part of the upstream failure set.

### DIFF
--- a/src-tauri/src/acp/connection.rs
+++ b/src-tauri/src/acp/connection.rs
@@ -1,6 +1,8 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::time::Duration;
 
 use sacp::schema::{
     BlobResourceContents, CancelNotification, ClientCapabilities, ContentBlock, ContentChunk,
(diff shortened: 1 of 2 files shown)


asxuen:
Could not push the macOS CI fix onto this fork head (`Permission denied` for asxuen → keh4l/codeg).

Root cause of the red macOS checks: tests spawn `/bin/true`, which does not exist on macOS (`/usr/bin/true` does).

Fixed PR with the same change set + path fix: https://github.com/xintaofei/codeg/pull/384

PR #441, opened 2026-08-11
fix(i18n): translate 61 remaining English strings in French locale

Charlie-1909:
## Summary

The French locale is structurally complete (4079/4079 keys, none missing), but 61 values were left in English and surface in the UI.

The largest gap is the **Remote Workspace** module: none of its 28 strings had been translated, including its dialogs and error toasts.

## Changes

- **RemoteWorkspace (28 strings)** — the entire module was untranslated
- **Conversation context bar (12)** — toasts and empty states (`No branches`, `Failed to open folder`, `Switched to {name}`, …)
- **Git vocabulary consistency (3)** — `commit`, `merge` and `push` in the context bar kept their English form, while the rest of the file already uses *Valider*, *Fusionner*, *Pousser*
- **Prompt terminology (2)** — `Automations.sectionPrompt` read "Prompt" while `Automations.prompt` already reads "Instruction"
- **Agent connection screen (2)** — `Initializing {agent} session`, `Creating session and loading configuration`
- **Theme colours (8)** and **misc (6)** — `Server ID`, `Save config.toml`, `Run Everything (--force)`, placeholders

## Deliberately left in English

Proper nouns (Git, HEAD, MCP, Stdio, Lark, shadcn), git CLI flags (`--hard`, `--soft`, `--keep`), `SKILL.md` templates (read by agents, which work in English), input placeholders (`ghp_xxx`, `pnpm install`, URLs), and words identical in French (Agent, Interface, Terminal, Configuration, …).

The project's existing glossary is preserved: **`tokens`** and **`Skill`** are kept untranslated, matching "Consommation de tokens" and "Sélectionnez une Skill" already present in the file.

## Verification

- Each key was checked against its expected English value before being edited — nothing was applied blindly
- `JSON.stringify(JSON.parse(fr.json), null, 2)` reproduces the original file byte-for-byte, so the diff carries no reformatting noise
- Key count unchanged: **4079 before, 4079 after**
- Only `src/i18n/messages/fr.json` is touched

Happy to adjust any wording if a different French term is preferred.

### DIFF
--- a/src/i18n/messages/fr.json
+++ b/src/i18n/messages/fr.json
@@ -576,7 +576,7 @@
       "draftTitle": "Nouveau MCP",
       "draftDescription": "Renseignez un Server ID et une configuration pour créer un serveur MCP local.",
       "serverIdLabel": "ID du serveur",
-      "serverIdPlaceholder": "Server ID (ex. my-mcp)",
+      "serverIdPlaceholder": "ID du serveur (ex. my-mcp)",
       "typeHint": "Types pris en charge : stdio, http (alias : streamable-http, streamableHttp), sse. env n'est utilisé que par stdio ; pour les MCP distants, utilisez headers pour transporter les jetons d'authentification.",
       "envOnRemoteWarning": "env détecté sur un MCP distant. Seul stdio utilise env ; les MCP distants transportent les jetons d'authentification via headers, donc env sera ignoré à l'enregistrement."
     },
@@ -759,7 +759,7 @@
       "permissionModeLabel": "Mode d'autorisation",
       "permissionModeDefault": "Demander avant d'exécuter (par défaut)",
       "permissionModeForce": "Tout exécuter (--force)",
-      "permissionModeHint": "Run Everything lance les sessions avec --force : tous les appels d'outils sont autorisés automatiquement sauf les règles de refus, sans confirmation (une politique d'organisation peut le limiter aux règles d'autorisation). S'applique aux nouvelles sessions.",
+      "permissionModeHint": "Tout exécuter lance les sessions avec --force : tous les appels d'outils sont autorisés automatiquement sauf les règles de refus, sans confirmation (une politique d'organisation peut le limiter aux règles d'autorisation). S'applique aux nouvelles sessions.",
       "optionDefault": "Défaut (non défini)",
       "sandboxLabel": "Bac à sable",
       "sandboxEnabled": "Activé",
@@ -3117,7 +3117,7 @@
         "other": "Autre",
         "otherPlaceholder": "Saisissez votre réponse…",
         "singleSelect": "Unique",
-        "multiSelect": "Choix multiple",
+        "multiSelect": "Multiple",
         "skip": "Ignorer",
         "next": "Suivant",
         "submit": "Envoyer",
@@ -3209,8 +3209,8 @@
     "conversationContextBar": {
       "folderTitle": "Dossier de travail",
       "branchTitle": "Branche de travail",
-      "searchFolder": "Rechercher un dossier…",
-      "searchBranch": "Rechercher une branche…",
+      "searchFolder": "Rechercher un dossier...",
+      "searchBranch": "Rechercher une branche...",
       "noFolders": "Aucun dossier",
       "noBranches": "Aucune branche",
       "noBranch": "(aucune branche)",
@@ -3220,10 +3220,10 @@
       "merge": "Fusionner",
       "toasts": {
         "folderChanged": "Basculé vers {name}",
-        "openFolderFailed": "Échec de l’ouverture du dossier",
+        "openFolderFailed": "Échec de l'ouverture du dossier",
         "switchedToChatMode": "Basculé en mode discussion",
-        "openStashFailed": "Échec de l’ouverture de la fenêtre de remisage",
-        "openMergeFailed": "Échec de l’ouverture de la fenêtre de fusion"
+        "openStashFailed": "Échec de l'ouverture de la fenêtre de remisage",
+        "openMergeFailed": "Échec de l'ouverture de la fenêtre de fusion"
       }
     },
     "cloneDialog": {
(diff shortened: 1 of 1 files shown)


xintaofei:
Thanks for this — it's a genuinely careful piece of work, and the PR description made it easy to review. 🙏

I went through it end to end and everything checks out:

- **No structural drift** — 4079 keys before and after, 0 added, 0 removed, exactly 61 values changed. Full key parity with `en.json` still holds and `src/i18n/messages.test.ts` passes.
- **Every one of the 61 old values was byte-identical to its `en.json` counterpart**, so nothing that had already been translated got quietly reworded.
- **ICU placeholders intact** — `{agent}`, `{name}`, `{message}` all preserved, checked against both the old French value and the English source. No unbalanced braces anywhere in the file.
- Nice touch writing the apostrophes as `’` rather than `'` — that also keeps them clear of ICU's quoting rules, since a straight `'` directly before a `{` would escape the placeholder.
- The round-trip claim holds: `prettier --check` is clean and the file reproduces byte-for-byte, so the diff really is free of reformatting noise.
- Your glossary claims all hold up against the existing file — `CommitPage.title` was already *Valider*, `branchDropdown.pushBranch` already *Pousser*, `Tasks.actionMerge` already *Fusionner*, `Automations.prompt` already *Instruction*. And `openStashFailed` matches the existing `branchDropdown.toasts.openStashWindowFailed` word for word. That level of checking is appreciated.
- The branch is a little behind `main` now, but it merges cleanly and the merged result keeps full parity (4132/4132).

**I'll merge this shortly.**

A few small things for a future pass — none are blockers, and there's no need to touch this PR:

1. `AcpAgentSettings.cursor.permissionModeHint` still opens with "**Run Everything** lance les sessions avec `--force`", while the option right above it now reads "Tout exécuter (--force)" — so the hint names a label that no longer appears in the UI. (`Tasks.settingsPromptsIntro` has a milder version of the same thing: the tab is now "Instructions" while the body text still says "prompt".)
2. `RemoteWorkspace.dragSort` — the file's other drag handles use the shorter "Glisser pour …" form, and `AcpAgentSettings.actions.dragSort` is already exactly "Glisser pour réordonner".
3. `askQuestion.multiSelect` → "Choix multiple" reads a little asymmetric beside its sibling `singleSelect` = "Unique" (they share the same badge). "Choix unique" / "Choix multiple" as a pair would be tidier.
4. `McpSettings.local.serverIdLabel` is now "ID du serveur" while `serverIdPlaceholder` directly below it still reads "Server ID (ex. my-mcp)".
5. More of a heads-up than a request: the theme palette names (`Slate`, `Stone`, …) are currently English in all ten locales, so `fr` becomes the first to translate them. I'm happy with *Ardoise* / *Pierre* — just flagging it in case you'd like to bring the other locales along later.

Thanks again — RemoteWorkspace had been sitting untranslated in most locales, so this is a real improvement.

xintaofei:
Rather than hand the follow-ups back to you, I've pushed them onto this branch myself — two commits on top of yours. Happy to drop or adjust any of it.

Before writing anything I ran two sweeps, and they changed a couple of my earlier suggestions:

**Cross-locale check.** Every locale renders the ask-question badge as a *symmetric pair* — `Single`/`Multiple` (en), `Única`/`Múltiple` (es), `Única`/`Múltipla` (pt), `Einfach`/`Mehrfach` (de), `单选`/`多选`, `단일 선택`/`다중 선택`. French was already `Unique`/`Multiple`, so I put `multiSelect` back to **"Multiple"** to restore that pair rather than promoting both keys. Similarly, `de` and `ar` translate the MCP label *and* its placeholder together, so `serverIdPlaceholder` is now "ID du serveur (ex. my-mcp)".

**"Same English → different French" sweep over the whole file.** This one was more useful, and it corrected me: several strings here had an exact English twin elsewhere in `fr.json` that was already translated. They're now byte-identical to those twins:

| key | was | now | twin |
|---|---|---|---|
| `RemoteWorkspace.dragSort` | Faire glisser pour réordonner | **Glisser pour trier** | `commandDropdown.manageDialog.dragSort` |
| `conversationContextBar.searchBranch` | …branche**…** | …branche**...** | `gitLogTab.branchSelector.searchBranch` |
| `toasts.openFolderFailed` / `openStashFailed` | l**’**ouverture | l**'**ouverture | 5 existing sites |

So my earlier "Glisser pour réordonner" suggestion was actually wrong — that's the translation of *"Drag to reorder"*, a different English string. "Drag to sort" already had "Glisser pour trier". Good catch on your part that I nearly got backwards.

`permissionModeHint` now opens with "Tout exécuter" to match the option above it; `--force` stays in both strings as the identifier.

**Things I deliberately left alone**, so they're not silently dropped:

- **The theme palette names stay translated.** I checked `appearance-settings.tsx` — every option renders a real colour swatch beside the label, so the label is decorative rather than an identifier. *Ardoise* / *Pierre* are good. I didn't push them into the other 9 locales; that's a separate, per-language call.
- **`settingsTabPrompts` = "Instructions" is correct.** The English source makes the same contrast itself — "Each stage already sends a built-in **prompt** … appended at the end as extra **instructions**" — so French mirrors it exactly.
- **`Merge` → "Fusionner" vs `Tasks.phaseMerge` → "Fusion"** is a verb/noun distinction, not a defect.
- **111 other same-English/different-French pairs** are genuine context-dependent wording — `None` → *Aucune*/*Aucun* (gender), `Remove` → *Supprimer*/*Retirer*, `Preview` → *Aperçu*/*Écouter* (the latter for a sound). Normalising those would make the translation worse.

The second commit is separate on purpose: it normalises three *pre-existing* strings whose two French renderings differed only by a curly vs straight apostrophe (`Failed to open folder`, `Auth Mode`, `Import failed: {message}`). It's unrelated to your translation pass, so drop that commit alone if you'd rather keep this PR strictly scoped. After it, no two identical English strings map to French strings differing only by punctuation — 0 exceptions, down from 3.

Verification on the final state: key parity 4079/4079 with `en.json`, no keys added or removed, ICU placeholders intact, no straight apostrophe landing before a `{` (I introduced several, so I re-checked), byte-for-byte round-trip, `prettier --check` clean, and the full frontend suite green locally — eslint, `tsc --noEmit`, 3793 tests across 290 files, and `next build`.

Thanks again for the original pass — the twin-matching sweep only found anything because your translations were close enough to compare.

PR #462, opened 2026-08-14
fix(grok): send images as native ACP blocks

Nothing-129:
## What changed

- Override Grok's advertised prompt image capability so pasted and attached images are sent as native ACP `image` blocks.
- Promote legacy queued image resources to native image blocks before sending them to Grok.
- Surface Grok's `image_dropped` notifications as non-terminal errors.
- Parse native Grok image chunks in history while preserving compatibility with the older resource-blob transcript shape.

## Why

Grok 1.0.2 still advertises `image: false`, so Codeg encoded images as embedded resources. Grok treats that shape as a binary file attachment and does not invoke its image-description sidecar, leaving the model unable to inspect the pixels. Native ACP image blocks do invoke the sidecar.

## Impact

Grok can now inspect pasted or attached images in Codeg. Existing queued drafts and older transcripts remain compatible, and users receive a visible warning when Grok rejects an image before sending it to the model.

## Validation

- `cargo test --features test-utils promote_grok_image_resources_lifts_image_blobs_only`
- `cargo test --features test-utils map_grok_ext_notification_image_dropped_surfaces_error`
- `cargo test --features test-utils merges_prompt_text_and_native_image_into_one_user_turn`
- `cargo test --features test-utils merges_prompt_text_and_image_resource_into_one_user_turn`
- `git diff --check`

All targeted tests passed.

### DIFF
--- a/src-tauri/src/acp/connection.rs
+++ b/src-tauri/src/acp/connection.rs
@@ -2523,20 +2523,34 @@ async fn apply_and_emit_session_config_options(
     emit_session_config_options_values(state, emitter, agent_type, updated).await;
 }
 
+/// Grok's initialize still advertises `image: false` — the coding model
+/// cannot see pixels. Native ACP `image` blocks nevertheless run its
+/// image-describe sidecar (verified against grok 1.0.2). An image-mime
+/// `resource` blob does not: grok dumps it as
+/// `<file_contents type="binary">` and the model only gets a path.
+/// Advertise `image: true` so the composer sends Image blocks.
+fn effective_prompt_capabilities(
+    agent_type: AgentType,
+    capabilities: &sacp::schema::PromptCapabilities,
+) -> PromptCapabilitiesInfo {
+    PromptCapabilitiesInfo {
+        image: capabilities.image || agent_type == AgentType::Grok,
+        audio: capabilities.audio,
+        embedded_context: capabilities.embedded_context,
+    }
+}
+
 async fn emit_prompt_capabilities(
     state: &Arc<RwLock<SessionState>>,
     emitter: &EventEmitter,
     capabilities: &sacp::schema::PromptCapabilities,
+    agent_type: AgentType,
 ) {
     emit_with_state(
         state,
         emitter,
         AcpEvent::PromptCapabilities {
-            prompt_capabilities: PromptCapabilitiesInfo {
-                image: capabilities.image,
-                audio: capabilities.audio,
-                embedded_context: capabilities.embedded_context,
-            },
+            prompt_capabilities: effective_prompt_capabilities(agent_type, capabilities),
         },
     )
     .await;
@@ -3626,6 +3640,7 @@ async fn run_connection(
                 &state,
                 &emitter_clone,
                 &init_resp.agent_capabilities.prompt_capabilities,
+                agent_type,
             )
             .await;
 
@@ -5726,6 +5741,31 @@ async fn journal_turn_span(
     let _ = tokio::time::timeout(std::time::Duration::from_secs(2), ack).await;
 }
 
+/// Promote image-mime embedded resources to native Image blocks.
+///
+/// Codeg used to send Grok images as `resource` blobs because grok
+/// advertised `image:false`. That path never reaches grok's describe
+/// sidecar — see [`effective_prompt_capabilities`]. Queued drafts and
+/// work-task replays can still carry the old shape; lift them here.
+fn promote_grok_image_resources(blocks: Vec<PromptInputBlock>) -> Vec<PromptInputBlock> {
+    blocks
+        .into_iter()
+        .map(|block| match block {
+            PromptInputBlock::Resource {
+                uri,
+                mime_type: Some(mime),
+                text: None,
+                blob: Some(blob),
+            } if mime.starts_with("image/") && !blob.is_empty() => PromptInputBlock::Image {
+                data: blob,
+                mime_type: mime,
+                uri: Some(uri),
+            },
+            other => other,
+        })
+        .collect()
+}
+
 fn map_prompt_blocks(blocks: Vec<PromptInputBlock>) -> Vec<ContentBlock> {
     blocks
         .into_iter()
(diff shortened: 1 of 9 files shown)


xintaofei:
Thanks for this one — I went through it carefully and it holds up really well. Rather than take the capability claim on faith I re-verified the premise live against the grok CLI on my machine, and everything you describe reproduces exactly.

## What I measured (grok 1.0.0 — the version `registry.rs` actually pins)

- `initialize` still advertises `promptCapabilities={image:false, audio:false, embeddedContext:true}`, yet a native ACP `image` block is accepted with no error at all.
- 64×64 PNG, four stripes green/orange/purple/black, asking for the colors top to bottom:
  - native image block → `green, orange, black, purple` (1409 reasoning tokens, 22s)
  - legacy embedded resource blob → `red, white, blue, green`, i.e. pure hallucination (5645 reasoning tokens, 70s)

  So the old path wasn't merely failing to show the image — it was burning ~4× the reasoning tokens guessing at it.
- Same image at 6000×6000, native block → `green, orange, purple, black`, exactly right, and **no** `image_dropped` fires (grok downscales silently). Good news for the new error arm: no false-positive banner for merely re-encoded images.
- 1×1 PNG → the real payload is `_x.ai/session_notification` with `{"sessionUpdate":"image_dropped","notes":["Image 1 was dropped before send: too small (1×1); images must be at least 8×8 pixels."]}` — byte-identical to your fixture, and that method is in `GROK_EXT_UPDATE_METHODS`, so the new arm genuinely fires.
- grok's on-disk `updates.jsonl` records the echo as `user_message_chunk` / `content:{type:"image",...}` with `_meta.promptIndex:0` on *both* chunks — precisely what the new parser test asserts.
- Bonus data point: the older 0.2.112 binary also accepts native image blocks and reads the pixels, so the un-version-gated override looks safe backwards as well.

## On the design

Putting the promotion at the `ConnectionCommand::Prompt` choke point is the right call. `map_prompt_blocks` has a single call site and every producer (chat, work-task engine, delegation, queued drafts) funnels through it, so nothing can smuggle the old shape past it — `reencode_images` in the work-task engine quietly becomes a no-op for grok once both caps are true, and the connection-level lift covers it anyway. And because every capability consumer reads `SessionState.prompt_capabilities` (populated from the emitted, already-overridden event), there's no split-brain between what the composer encodes and what the dispatcher sends. Nice side effect too: `grok_ext_notification_is_turn_output` picks up `image_dropped` for free, so a drop-only turn stops misreporting as `turn_failed_empty`.

## Checks run on the branch

- `cargo clippy --all-targets --features test-utils -- -D warnings` — clean
- `cargo test --features test-utils` — 2455 passed / 0 failed
- `cargo clippy --no-default-features --bin codeg-server --lib -- -D warnings` — clean
- `cargo test --no-default-features --bin codeg-server --lib` — 2431 passed / 0 failed

## Optional nits (nothing blocking)

1. The banner reads a little doubled: grok's note already opens with "Image 1 was dropped before send…", so the UI ends up showing `Image dropped: Image 1 was dropped before send: too small (1×1)…`.
2. Three comments now describe the old world — `message-input-attachments.ts:55-65` still says Grok takes the embedded-resource shape and "the sent payload is unchanged for them", plus the `reencode_images` doc in `work_task/engine.rs` and the `UserMessageBlock` doc at `acp/types.rs:474`. The paths they describe are still needed for old drafts/transcripts, so it's just wording drift.
3. The PR body and the new doc comment say "verified against grok 1.0.2", but `registry.rs` pins 1.0.0. Worth naming the pinned version so a future reader doesn't assume the behavior is 1.0.2-only — I checked, 1.0.0 behaves identically.
4. `session/load` replay runs ext notifications through the same mapper, so an old session replayed via that fallback could resurface a historical image-drop alert. Same shape as the existing `auto_compact_failed` arm and non-terminal, so cosmetic.
5. `image/svg+xml` / `.ico` attachments will now be promoted and then rejected by grok's normalizer with a visible error, instead of silently riding along as a binary dump. Arguably an improvement — just flagging it as a user-visible change.

Looks good to me — I'll merge this shortly. Thanks for chasing the actual wire behavior instead of trusting the advertised capability; that's the part that made this fixable at all.

xintaofei:
Pushed `45d04133` — the five nits from my last comment are all closed. One of them turned out to be more than a nit, so here's what happened.

## The SVG note was an actual regression

I flagged `image/svg+xml` last time as "arguably an improvement, just flagging it". That was wrong — I went and measured it:

| `image/svg+xml`, asking for the 4 stripe colors | model's answer |
|---|---|
| resource blob (before this PR) | **green, orange, purple, black** — exactly right |
| native image block (after this PR) | dropped, then "the image could not be loaded" |

Grok's validator rejects svg outright (`image_dropped`: "validate: unsupported or unrecognised image format"), so the model gets nothing. Carried as a resource blob the file lands in the session's `assets/` and the model just reads the source — SVG is text, so it wins that round. My earlier "visible error beats silent hallucination" reasoning missed that the old path wasn't failing at all here.

The composer only sees one `image` capability bit, so it can't make this call per mime. Fix is at the same choke point you already picked: `promote_grok_image_resources` → `normalize_grok_image_blocks`, now sorting both directions — promote what grok decodes, demote what it doesn't. The allow-list is grok's own raster set (`image/{png,jpeg,gif,webp,bmp,tiff}`); I verified png, webp, bmp and tiff each round-trip through the describe sidecar with correct answers, and that anything off the list keeps the pre-PR carriage, so a wrong entry there can only ever be conservative.

## The other four

- **Message stutter** — grok's note already opens with "Image 1 was dropped before send…", so it's shown verbatim now. Only the shapeless `reason`/`message` fallbacks get a prefix, and a blank one no longer renders as a bare `Image dropped: `.
- **Stale alerts on replay** — this one bit harder than expected once I looked: the `session/load` replay closure forwards ext notifications unconditionally, so a *past* session's dropped image (or failed compaction, which predates this PR) came back as a live status-bar alert **plus an OS notification**, just for opening the conversation. Added `grok_ext_notification_is_alert`, mirroring your `grok_ext_notification_is_turn_output`, and the replay now skips those — the typed closure right beside it already drew the same line by forwarding only `AvailableCommandsUpdate`. Compaction *cards* still replay.
- **Version claim** — the doc now cites 1.0.0 (what `registry.rs` pins) and 0.2.112, and says why it's deliberately not version-gated.
- **Stale comments** — cleaned up in `acp/types.rs`, `work_task/engine.rs`, `message-input-attachments.ts`, plus a few more that still used Grok as the example of an agent receiving images as embedded resources (`use-composer-attachments.ts`, `message-input.tsx`, `prompt-draft.ts`, `from-prompt-blocks.ts`) and the two mapper doc comments that only mentioned compaction.

Four new tests cover the demotion (including an upper-cased mime, since the two guards have to read the same string the same way), the allow-list boundary, the note-less drop, and the alert predicate.

## Checks

- `cargo clippy --all-targets --features test-utils -- -D warnings` — clean
- `cargo test --features test-utils` — 2459 passed / 0 failed
- server mode: clippy clean, `cargo test --no-default-features --bin codeg-server --lib` — 2435 passed / 0 failed
- `npx tsc --noEmit`, `pnpm eslint src` — clean
- `pnpm test` — 3829 passed; `pnpm build` — green

Thanks again for the original find — the `image:false` advertisement being simply untrue is the kind of thing nobody catches without actually poking the wire.

PR #484, opened 2026-08-16
feat(settings): Ctrl/Cmd + and - step the window zoom

Adam-Dalloul:
Settings → Window zoom is a discrete menu. This binds the usual keys to those same rungs:

- Ctrl/Cmd + or = zooms in one step
- Ctrl/Cmd - zooms out one step
- Ctrl/Cmd 0 resets to 100%

Rungs are 80 / 90 / 100 / 110 / 125 / 150 / 175 / 200 / 250 / 300. Capture-phase so the webview does not eat the keys as page zoom. Same persist path as the Settings menu.

Also listed under Settings → Shortcuts.


xintaofei:
Thanks for this — ⌘/Ctrl `+` `-` `0` is a genuinely missing affordance, and I like that you reused the existing discrete rungs instead of inventing a second, continuous zoom model. Putting the listener in `AppearanceProvider` is also the right call: it's mounted from the root layout, so it covers the settings and pet windows and web mode, whereas `WorkspaceChromeController` would have missed them. Extracting `stepZoom` with a unit test and remembering to keep `VALID_ZOOMS` in the pre-paint inline script in sync are both nice touches — that second one is easy to miss.

I ran the branch: `tsc --noEmit`, ESLint and Prettier are clean, and the full suite is green (313 files / 4198 tests). Hygiene is good.

There's one thing I'd like fixed before merging, plus a few smaller ones. Details below.

---

## Blocking

### 1. "Zoom in" / "Zoom out" show up as rebindable in Settings, but nothing reads the binding

`zoom_in` / `zoom_out` / `zoom_reset` are added to `SHORTCUT_DEFINITIONS` and `DEFAULT_SHORTCUTS`, so `ShortcutSettings` renders three rows with the full record / conflict-check / reset-to-default UI. But at runtime only `zoom_reset` is actually consulted (`appearance-provider.tsx:754`, `:782`) — zoom in/out go through the hardcoded `isZoomInShortcutEvent` / `isZoomOutShortcutEvent`, which take no binding argument (`appearance-provider.tsx:772`, `:777`).

So if a user rebinds "Zoom in" to, say, ⌘⇧Z: the value is normalized, persisted, the button label updates and a success toast fires — and then nothing happens when they press it, while ⌘= keeps zooming. Worse, rebinding frees `mod+=` in the settings conflict checker, so they can now legitimately assign ⌘= to another action and get **both** behaviors (the zoom handler only calls `preventDefault()`, not `stopPropagation()`).

Two ways out, either is fine:

- **Wire them up:** match against `shortcuts.zoom_in` / `shortcuts.zoom_out` like every other action. If you want to keep the shift/numpad tolerance, put it in the matcher for the *bound* key (treat `=`↔`+` and `-`↔`_` as the same physical key) rather than in a separate predicate that bypasses the binding.
- **Or make them honest:** drop `zoom_in` / `zoom_out` from `SHORTCUT_DEFINITIONS`, keep them fixed, and just document them. `zoom_reset` is already wired correctly and can stay.

### 2. `if (event.repeat) return` runs *before* the match, so held keys aren't prevented

`appearance-provider.tsx:771`. Two consequences:

- Holding ⌘- doesn't repeat, so it's 7 taps to walk 100% → 300%. Every other app repeats here.
- In server/browser mode (`codeg-server` / Docker / remote access) the un-prevented repeat events reach the browser's *own* zoom, which then compounds on top of the app's rem zoom.

Moving the repeat guard after the match — and still calling `preventDefault()` — fixes both. Or just allow repeats.

### 3. The terminal loses ⌃`_` (readline/emacs `undo`)

The listener is a bare `window` capture handler with no target check, and `mod` in this codebase means Ctrl **or** Cmd (`hasMod = event.metaKey || event.ctrlKey`). `isZoomOutShortcutEvent` also deliberately ignores `shiftKey`, so ⌃⇧`-` (= ⌃`_`) matches too.

Net effect with the integrated terminal focused — on macOS as well as Linux/Windows — pressing ⌃`_` sends `undo` to readline *and* steps the app zoom. Both fire.

`terminal-context.tsx` already has the pattern for this (`isInTerminalRegion`, `terminal-context.tsx:304`); reusing it here would keep terminal muscle memory intact. I know ⌃K-vs-`toggle_search` has the same shape today, so this isn't a regression you introduced — but zoom keys collide with a lot more shell muscle memory than ⌘K does, so I'd rather not widen it.

### 4. Recording *any* shortcut in Settings also zooms the UI

The recorder in `shortcut-settings.tsx:59` calls `event.stopPropagation()`, but both it and the zoom handler are capture listeners on the **same** `window` node — `stopPropagation()` doesn't stop sibling listeners on the same target. So while the "Record" button is armed and the user presses ⌘= or ⌘-, the whole settings window zooms underneath them.

(And per #5 below, ⌘⇧= is then rejected as invalid anyway, so they get a zoom plus an error toast.)

### 5. `+` can't actually be recorded

`shortcutFromKeyboardEvent` joins parts with `"+"`, so ⌘⇧= serializes to `"mod+shift++"`; `normalizeShortcut` splits on `"+"` and returns `null`, so `updateShortcut` fails and the user gets "invalid shortcut". `normalizeShortcut("mod++") === null` — I checked.

This one predates the PR, but the PR is what puts a `+`-shaped row in front of users, so it becomes reachable here.

### 6. i18n: the 9 non-English locales got English strings

`zoom_in` / `zoom_out` / `zoom_reset` titles and descriptions are English in ar, de, es, fr, ja, ko, pt, zh-CN and zh-TW. The convention here is that new UI strings ship translated in all ten (cf. `toggle_custom_style` in 4e547089).

Separately, `Appearance.zoomLevel.sectionDescription` was only extended in `en.json`, so the shortcut hint is invisible in every other language. Two copy notes on the English while you're in there: it says "Ctrl" only, and on macOS it's ⌘; and "one Settings rung" is a bit opaque as user-facing text.

---

## Non-blocking

- **Split the 300% commit.** `feat(settings): raise window zoom to 300%` widens `ZOOM_LEVELS` from 6 rungs to 10 and is an independent product decision from the keybinding. I checked and it's less risky than I first assumed — Monaco (`file-workspace-panel.tsx:2291`) and xterm (`terminal-view.tsx:456`) both scale their font size off `zoomLevel`, and the Rust side already clamps 50–300 — but I'd still rather see it land on its own with a screenshot at 300%.
- **The `event.code === "Minus" / "Equal"` fallback misfires on non-US layouts.** `{ key: ")", code: "Minus", ctrlKey: true }` returns zoom-out (that's the AZERTY shape). The `event.key` branch already covers the real `-` / `+` / `=` on those layouts, so I'd narrow the `code` fallback to the numpad keys or drop it.
- **`matchShortcutEvent(event, "mod++")` in `isZoomInShortcutEvent` is dead** — `normalizeShortcut("mod++")` is `null`, so it's always false. Harmless (the fallback catches `+`), but it means the ⌃⇧`+` test is passing through a different branch than it looks like.
- **The three new `SPECIAL_KEY_ALIASES`** (`keyboard-shortcuts.ts:135-137`) are inert on the webviews we ship on: `normalizeKeyToken` returns early for single characters, and Chromium/WebKit report `"+"` / `"-"` / `"="`, never `"Add"` / `"Subtract"` / `"Equal"` (`"Equal"` is a `code`, never a `key`). And `normalizeShortcut("mod+add")` → `"mod++"`, which the parser then rejects — so the round-trip is broken. I'd drop them, or fix them along with #5.
- **`zoomLevelRef.current = zoomLevel` in the render body** (`appearance-provider.tsx:481`). No demonstrated bug, but `terminal-view.tsx:449` does the same sync inside an effect — worth matching.
- **No `isComposing` guard** on the global listener, so an IME's own Ctrl/Cmd `+`/`-` commands can be intercepted. Probably falls out of whatever target policy #3 gets.
- **Tests.** `keyboard-shortcuts.test.ts:134` and `:138` are the same assertion twice. More useful coverage would be: the configurable path (rebind → new binding fires, old one doesn't), `repeat` events, and one non-US-layout event.
- **Docs.** The canonical shortcut table lives in the docs site (`guide/workspace.md` and its `zh/` twin), so the three new rows want a companion PR there.

---

None of this is structural — the shape of the change is right, it's mostly about making the Settings rows tell the truth and picking a target policy for the keys. Happy to look again once #1 is sorted; the rest can come along with it or as follow-ups if you'd prefer. Thanks again for taking this on! 🙏

Adam-Dalloul:
Thanks for catching that. The Settings rows for zoom in/out were decorative, which is on me.

Follow-up is up: those bindings actually drive the listener now (`=` / `+` and `-` / `_` stay equivalent on the bound key). Held keys repeat and still preventDefault so the browser doesn't stack its own zoom. Terminal keeps Ctrl+_, recording a shortcut no longer zooms the settings window, and `+` can actually be recorded. Strings are translated, and the zoom hint no longer says Ctrl-only.

Also dropped the dead Add/Subtract/Equal aliases and limited the code fallback to the numpad so AZERTY Ctrl+) doesn't zoom out.

300% is still its own commit here. Happy to split it if you'd rather land that separately.

Adam-Dalloul:
Thanks for catching that. The Settings rows for zoom in/out weren't actually wired, which is on me.

Follow-up is up. Those bindings drive the listener now, and = / + and - / _ still count as the same key. Held keys repeat and still preventDefault so the browser doesn't also zoom. Terminal keeps Ctrl+_, recording a shortcut no longer zooms the settings window, and + can actually be recorded. Strings are translated, and the zoom hint no longer says Ctrl-only.

Also dropped the dead Add/Subtract/Equal aliases and limited the code fallback to the numpad so AZERTY Ctrl+) doesn't zoom out.

300% is still its own commit here. Happy to split it if you'd rather land that separately.

xintaofei:
Thanks for the thorough turnaround — this is a much better change than what I reviewed last week, and you fixed more than I asked for.

I re-ran everything on `ae6982b4` and went through each item:

| Round-1 item | Status |
| --- | --- |
| 1. Settings rows not wired | ✅ `resolveWindowZoomAction` drives off the live bindings. Verified: remapping `zoom_in` to `mod+shift+z` makes ⌃⇧Z fire and the old ⌃= go dead. |
| 2. `repeat` returned before `preventDefault` | ✅ Match first, prevent unconditionally. |
| 3. Terminal lost ⌃`_` | ✅ Real guard — `[data-terminal-panel-region="true"]` is genuinely rendered (`terminal-panel.tsx:12`) and is the same selector `isInTerminalRegion` uses. |
| 4. Recording also zoomed | ✅ And you got the subtle part right: `stopImmediatePropagation()` alone would have been too late, since the provider's listener is registered first. The `shortcutRecorderArmed` flag is what actually does it. |
| 5. `+` unrecordable | ✅ `parseShortcut`'s trailing-`+` handling is clean. `mod+shift++` round-trips, matches its own event, and renders as `⌘⇧+` / `Ctrl+Shift++`. |
| 6. i18n | ✅ All ten locales translated, `sectionDescription` mirrored everywhere, ⌘/Ctrl wording fixed. |

Plus the nits: AZERTY `code` false positive gone (narrowing to `NumpadAdd`/`NumpadSubtract` was the right call), dead `mod++` branch and dead aliases removed, ref moved into an effect, `isComposing` guard added, duplicate assertion gone.

Since `matchShortcutEvent` / `normalizeShortcut` are shared by *every* shortcut, I probed the rewrite for collateral damage: all 19 defaults still match their canonical events and normalize idempotently, and surplus/missing Shift is still exact for non-sibling keys (⌘⇧K doesn't match `mod+k`, Shift+Enter doesn't match `enter`). Generated canonical legacy strings round-trip fine, so `parseShortcut` itself carries no migration risk. Full suite green (4718), ESLint and Prettier clean.

Three things before merge, one of which is new and I think matters most.

---

## 1. New default bindings silently double-bind over existing users' customizations

This is the one I'd most like fixed, and it's my fault for not seeing it in round 1.

`normalizeSettings` seeds `{...DEFAULT_SHORTCUTS}` and then overlays the persisted subset, so the three new zoom defaults land unconditionally — including on top of a chord the user already assigned to something else. There's no load-time conflict handling; the only conflict check is the string comparison at record time (`shortcut-settings.tsx:80`).

I confirmed it with a payload the *currently shipped* app could have written:

```js
localStorage["settings:shortcuts:v1"] = '{"toggle_search":"mod+-","toggle_sidebar":"mod+0"}'
readShortcutSettings()
// → toggle_search: "mod+-"   zoom_out:   "mod+-"    ← same chord
// → toggle_sidebar: "mod+0"  zoom_reset: "mod+0"    ← same chord

// ⌘- now matches toggle_search AND resolves to zoom "out"
// ⌘0 now matches toggle_sidebar AND resolves to zoom "reset"
```

So anyone who'd bound an action to ⌘-, ⌘= or ⌘0 gets both firing after upgrade, silently, with no way to discover it except by using the app. Simplest fix is to skip seeding a new default when a persisted binding already claims that chord (using matcher semantics, not string equality, per #3 below). This is really a gap in the shortcuts system rather than something you introduced — but this PR is the first to add new defaults since it shipped, so it's the first to hit it.

## 2. Held keys now write to SQLite on every repeat

Fixing the repeat guard was right, but `setZoomLevel` (`appearance-provider.tsx:474`) has no unchanged-value short-circuit, and it unconditionally calls `syncTrafficLightPosition` → Tauri IPC → `update_traffic_light_position` → `app_metadata_service::upsert_value`, which is a real on-disk SQLite write. The `#[cfg(target_os = "macos")]` gate only covers the atomic store; the DB write is outside it, so all three desktop OSes pay it (web/server mode escapes the IPC, only localStorage).

`stepZoom` clamps at both ends, so holding ⌘- at 80% emits one IPC + one SQLite upsert per key-repeat, indefinitely, with nothing changing on screen.

Worth noting the obvious fix isn't quite enough: guarding on `stepZoom` returning the current level covers in/out but not `zoom_reset`, which calls `setZoomLevel(DEFAULT_ZOOM_LEVEL)` unconditionally — holding ⌘0 storms too. A no-op guard covering all three (or an early return in `setZoomLevel` itself when the level is unchanged) handles both.

While you're in there: `zoomLevelRef` is now synced in a passive effect, so a fast repeat burst can read a ref lagging the state and drop a step. I haven't reproduced it — 33 ms repeat intervals should let passive effects flush — but a `stepZoomLevel(direction)` on the provider using `setZoomLevelState(prev => …)` would fix this and #2 in one move, and drop the ref entirely.

## 3. ⌘0 can't be pressed at all on AZERTY

`zoom_reset: "mod+0"` is the first digit-based default in this app, so nothing hit this before. On French/Belgian AZERTY the digit row is shifted — unshifted `Digit0` is `à`, and `0` requires Shift. Both spellings fail:

```
Cmd+Shift+0  → key "0", shiftKey true  → rejected by the surplus-Shift rule
Cmd+0        → key "à"                  → doesn't match "0"
```

Verified both return `null` from `resolveWindowZoomAction`. Zoom in/out are fine there, since `-` and `=` are unshifted on AZERTY — it's only reset that's unreachable.

`eventKeyToken` already knows this trick (it maps `Digit([0-9])` → the digit) but only when `altKey` is set. Extending the existing `code` fallback to accept `Digit<N>` when the bound key is a digit would mirror what you did for the numpad.

---

## Non-blocking

- **Conflict detection can now miss a real collision.** Settings compares binding strings exactly, but sibling-key + surplus-Shift means two *different* strings match the same event: `mod+=` and `mod+shift++` both match ⌘⇧= (verified), so Settings reports no conflict and both actions fire. Same mechanism inside zoom itself — an event matching two zoom bindings always resolves `in` first, so the other silently never fires. Comparing with matcher semantics instead of `===` fixes this and #1 together.
- **The terminal guard is region-wide**, so ⌘= over the terminal *tab bar* also declines. Correct for ⌃`_`; slightly surprising for ⌘= on macOS, where it has no shell meaning. Narrowing the bail to `event.ctrlKey && !event.metaKey` would keep both. (To be precise about the current behavior: the guard returns before `preventDefault()`, so the key still reaches xterm — it's only Codeg's zoom that declines.)
- **`toggle_custom_style` 20 lines above doesn't check `isShortcutRecorderArmed()`**, so recording a shortcut still fires ⌘⌥⇧S. Pre-existing, but now inconsistent with its neighbour.
- **Numpad keys still can't be *recorded*.** `matchShortcutEvent` accepts them via `code`, but `shortcutFromKeyboardEvent` returns `null` for a legacy `key: "Add"`, so the matcher supports a shape the recorder can't produce. Harmless asymmetry, just noting it.
- **`"still matches a repeat of the bound zoom key"`** never sets `repeat` — `resolveWindowZoomAction` doesn't read it. The repeat logic lives in the provider and is currently untested; a small test there would cover #2 and the endpoint behavior.
- **300%** — yes please, split it out. Happy to take it as a follow-up PR.

---

Really nice work on the fix commit — the `parseShortcut` rewrite in particular is cleaner than what was there before, and the recorder-flag reasoning was exactly right. #1 is the one I'd want addressed; #2 and #3 are small and localized. Ping me when it's up and I'll take another look. 🙏

Adam-Dalloul:
All three are in, force-pushed as `94b4302` (rebased onto main while I was there).

1. `normalizeSettings` now tracks which ids actually came from storage and skips seeding a default whose chord a stored binding already claims. Your payload gives `zoom_out: ""` and `zoom_reset: ""` with `zoom_in` untouched. The unbound state has to round-trip through `writeShortcutSettings` as well, otherwise the next unrelated settings change reseeds the collision, so an empty string is now a real stored value rather than a missing key. Settings renders those rows as Unassigned, and Reset defaults still brings them back. Shared pairs like `new_terminal_tab` / `new_conversation` are exempt, and two defaults colliding with each other is left alone.

2. `setZoomLevel` returns early when the level is unchanged, which covers reset at 100% as well as the clamped ends, so a held key stops reaching the IPC and the upsert entirely. Added `stepZoomLevel(direction)` on the provider. I kept a ref rather than the functional updater you suggested, but write it in the setter instead of a passive effect: with `setZoomLevelState(prev => ...)` the persist and IPC end up inside the updater, and StrictMode double-invokes that. The new test fires three repeats inside one `act()` so nothing flushes between them, and it fails if the ref goes back on an effect. Also covered the two endpoint cases and the preventDefault on repeat.

3. The `code` fallback now takes `Digit<N>` when the bound key is a digit, and a surplus Shift is tolerated when the event's code is that physical key, so `Cmd+0` and `Cmd+Shift+0` both reach reset on AZERTY. Scoped to the digit row, so a letter is still exact and Ctrl+`)` with code `Minus` is still not zoom out.

Took three of the non-blocking ones. Conflict detection compares with matcher semantics now: `shortcutsConflict` runs each binding's canonical event back through `matchShortcutEvent`, so `mod+=` against `mod+shift++` is caught, and that is the same predicate the seeding guard in 1 uses. Terminal guard is narrowed to `event.ctrlKey && !event.metaKey`, so Cmd+= over the terminal zooms again. And `toggle_custom_style` checks `isShortcutRecorderArmed()` like its neighbour. Left the numpad recording asymmetry as is.

tsc clean, eslint and prettier clean, full suite green at 4865.

300% I will send separately.

Adam-Dalloul:
Split the 300% commit out into #569 as requested, so this PR is now just the shortcut work. It stacks on this branch, easiest to take after this one lands.

xintaofei:
All three items from last round are properly fixed — I verified each against the code rather than the commit message, and the migration work in particular is exactly right.

| Round-2 item | Status |
| --- | --- |
| 1. New defaults double-binding over stored chords | ✅ Verified with the reported payload: `{"toggle_search":"mod+-","toggle_sidebar":"mod+0"}` now loads as `toggle_search: mod+-` / `zoom_out: ""` and `toggle_sidebar: mod+0` / `zoom_reset: ""`, with `zoom_in` untouched. Idempotent across three read→write→read cycles, the intentional `mod+t` / `mod+w` pairs survive, a fresh install still equals `DEFAULT_SHORTCUTS` exactly, and `""` is inert in the matcher rather than a wildcard. A stored *sibling* form (`mod+shift++`) also correctly drops `zoom_in`, which is the part that proves the conflict check is really being used. |
| 2. IPC + SQLite write per key repeat | ✅ `setZoomLevel` early-returns on an unchanged level and writes the ref synchronously; `stepZoomLevel` reads it; the cross-window storage path updates it too. Covers reset-held-at-100%, which was the half a naive `stepZoom`-changed guard would have missed. |
| 3. ⌘0 unreachable on AZERTY | ✅ Reachable both ways now — `Ctrl+0` (`à`) and `Ctrl+Shift+0` (`0`) both resolve to reset. |
| Terminal guard / recorder flag | ✅ `ctrlKey && !metaKey` keeps ⌘ zooming on macOS with the terminal focused while ⌃ defers to readline; `toggle_custom_style` checks the recorder flag like its neighbour. |
| i18n | ✅ `unassigned` in all ten locales, genuinely translated. |

`shortcutsConflict` round-tripping each binding through the *real* matcher via a synthetic event is nicer than what I had in mind — comparing against the matcher instead of re-deriving its rules means the two can't drift, and doing it bidirectionally is what catches `mod+=` / `mod+shift++`. Good call.

No regressions: all 19 actions still match their canonical events and normalize idempotently, and Shift stays discriminating for ordinary keys (⌘⇧K doesn't match `mod+k`, Shift+Enter doesn't match `enter`, ⌘Tab doesn't match `mod+shift+tab`). `tsc --noEmit` clean, ESLint clean, Prettier clean, full suite 353 files / 4865 tests green.

---

## One thing I'd like to fix, from the AZERTY approach

`matchesDigitRowCode` accepts `event.code === "Digit<N>"` whatever character came out, and the surplus-Shift tolerance is gated on that match. On a shifted digit row that makes one keypress satisfy two bindings — and `shortcutsConflict` structurally can't see it, because the recorder serialises from `event.key` while the matcher matched by `event.code`, and `syntheticEvent` only rebuilds the `key` form (attaching `code` only when the bound key is itself a digit). The two representations never meet.

Two consequences, both reproduced:

**US/UK — the double-fire is back, narrowly.** ⌘⇧0 arrives as `{key:")", code:"Digit0"}` and matches `mod+0`. Recording that chord for some other action stores `mod+shift+)`. `shortcutsConflict("mod+0", "mod+shift+)")` returns `false`, so Settings reports no conflict — and then one press runs that action *and* zoom-reset. Same class this commit exists to close.

**German QWERTZ — two shipped defaults overlap.** ⌃⇧0 produces `=`, so it matches `mod+0` by code *and* `mod+=` by character. A sweep over `DEFAULT_SHORTCUTS` found exactly one undeclared overlap: `zoom_in` + `zoom_reset`. It behaves acceptably today only because `resolveWindowZoomAction` happens to test `zoom_in` first, so `=` wins — which is the right answer for that keypress, but by luck of ordering rather than by design.

Two lines fix both. I applied them locally and your 31 shortcut tests still pass:

```ts
function matchesDigitRowCode(event: ShortcutEventLike, boundKey: string): boolean {
  if (!DIGIT_KEY_PATTERN.test(boundKey)) return false
  // Unshifted only: with Shift the digit row yields a character that belongs to
  // someone else — ")" on US, "=" on QWERTZ — and claiming it by position makes
  // one press satisfy two bindings.
  if (event.shiftKey) return false
  return event.code === `Digit${boundKey}`
}
```

and gate the surplus-Shift tolerance on the binding rather than on that match, so a digit binding stays Shift-agnostic on the *character* path (this is what keeps AZERTY's `Ctrl+Shift+0` spelling working):

```ts
} else if (event.shiftKey && keys.size === 1 && !DIGIT_KEY_PATTERN.test(parsed.key)) {
```

After that: QWERTZ ⌃⇧0 → only `mod+=` → in; US ⌘⇧0 → no longer touches `mod+0`, double-fire gone; AZERTY ⌃0 (`à`) → reset; AZERTY ⌃⇧0 (`0`) → reset; US ⌃0 → reset; and both conflict control cases still report true.

**Being straight about what that does and doesn't do:** it removes the shifted half, which is where the realistic case lives, but it narrows the class rather than closing it. The unshifted half remains — on AZERTY `-` sits on `Digit6`, so `Ctrl+-` matches both `mod+-` and `mod+6` with `shortcutsConflict("mod+6","mod+-")` returning false (same for `_` on `Digit8`). Nothing binds digits 1–9 today, so it's latent rather than live, but any `Digit<N>` positional fallback has this shape and it's worth a comment saying so.

**How much I care:** not enough to hold the merge. The US case needs someone to deliberately assign ⌘⇧0 elsewhere, the QWERTZ one resolves to a sensible answer, and nothing here risks data. It's two verified lines, so my preference is to land them in this PR while the context is fresh — but if you'd rather ship and follow up, that's fine by me and I'll take it as a separate PR.

## Non-blocking

- **Stored-vs-stored conflicts aren't repaired.** The seeded-default loop skips a pair where both ids came from storage, so a duplicate the old string-equality check let through survives the upgrade. Correct scope for this PR — just noting the new check only guards *future* additions.
- **300%** is still bundled. Happy to take it separately whenever; it's isolated in its own commit so it's easy either way.

Really good work across all three rounds — the migration handling especially is the kind of thing that's easy to wave off as "who would even have that binding", and you took it seriously.

Adam-Dalloul:
Landed your two lines in `1bc10ff8`. Shifted digit keys no longer match by position, so US Cmd+Shift+0 stops double-firing and QWERTZ Ctrl+Shift+0 resolves to zoom in only. The surplus-Shift tolerance is now gated on the binding being a digit, which keeps the AZERTY Ctrl+Shift+0 spelling working for reset.

I also wrote down the part you flagged as still open: the unshifted half is latent because AZERTY puts - on Digit6 and _ on Digit8, and shortcutsConflict cannot see it since the recorder serialises from key while this matches by code. That is a comment on the function now, so the next person hits it before rediscovering it. Two tests cover the shifted cases, full suite 353 files / 4867 green.

The 300% commit is already out of this branch, it went to #569 yesterday.

Adam-Dalloul:
Landed your two lines in `1bc10ff8`. Shifted digit keys no longer match by position, so US Cmd+Shift+0 stops double-firing and QWERTZ Ctrl+Shift+0 resolves to zoom in only. The surplus-Shift tolerance is now gated on the binding being a digit, which keeps the AZERTY Ctrl+Shift+0 spelling working for reset.

I also documented the half you flagged as still open, since AZERTY puts - on Digit6 and _ on Digit8 and shortcutsConflict cannot see that (the recorder serialises from key, this matches by code). It is written on the function itself rather than left for someone to work out again. Two tests cover the shifted cases, full suite 353 files / 4867 green.

The 300% commit is already out of this branch, it went to #569 yesterday.

xintaofei:
This is good to go from my side. 🎉

I verified the new commit does what it says rather than trusting the description:

- **US ⌘⇧0** no longer touches `mod+0`, so the double-fire is gone — recording that chord for another action and pressing it now runs exactly one thing.
- **QWERTZ ⌃⇧0** resolves to `in` only, matching `mod+=` and no longer also `mod+0`.
- **AZERTY still reaches reset both ways** — `Ctrl+0` (`à`) and `Ctrl+Shift+0` (`0`) both work, which was the whole point of the fallback and the part that was easy to break while narrowing it.
- A sweep over `DEFAULT_SHORTCUTS` against a layout-spanning event set now finds **zero** undeclared overlaps, down from the one (`zoom_in` + `zoom_reset`) I reported.
- No regressions: all 19 actions still match their canonical events, and Shift stays discriminating for ordinary keys.
- Full gate green: 4867 tests, `tsc --noEmit` clean, ESLint clean.

I especially appreciate that you wrote the residual into the comment instead of quietly leaving it — spelling out that the unshifted half remains, that AZERTY puts `-` on `Digit6` and `_` on `Digit8`, and that any `Digit<N>` positional fallback has this shape. That's the note that stops the next person from "simplifying" it back into a bug, and it's exactly the right way to close out a known limit.

Thank you for the patience across four rounds on this. Each time you came back with the fix *and* a test that pins the behaviour, and more than once you got a subtlety right that I hadn't spelled out — the `stopImmediatePropagation`-is-too-late reasoning behind the recorder flag, and the synthetic-event round-trip in `shortcutsConflict`, which is a genuinely better design than comparing bindings by hand. The migration handling in particular is the kind of thing most PRs would have waved off as "who would even have that binding", and you took it seriously and covered it properly.

Merging this. #569 review coming right up.

PR #487, opened 2026-08-16
fix(chat): stop leftover * and _ at the end of finished replies

### LINKED ISSUE #555: Markdown renderer draws the closing code fence inside the block when the message ends exactly at the fence
codeg v0.28.0, Windows 11, codex agent (but the bug is in the renderer, not the agent).

## Symptom

A message that ends with a fenced code block renders the CLOSING fence as a second content line inside the block, with the caret sitting after it:

```
1  e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
2  ```_
```

## The raw source is valid

I pulled the exact assistant text from the agent's session file on disk. The message tail is well-formed and ends immediately after the closing fence, with no trailing newline:

```
SHA-256:\n\n```text\ne3b0c442...52b855\n```
```

So the parser fails to terminate the block when the closing ``` is the final bytes of the message (no `\n` after it), and prints the backticks as code content instead. A block with any text following it in the same message renders fine.

Cosmetic only, but it makes copy-buttoned output (hashes, keys, one-liners) look wrong and the copied text can include the stray fence line.

Repro: have any agent reply with a message whose last characters are a closing fence, e.g. exactly:

~~~
SHA:

```text
abc123
```
~~~

Adam-Dalloul:
Finished assistant replies were showing a stray `*` or `_` at the end. Not the model. Streamdown defaults to `mode="streaming"`, so remend ran on every completed bubble. remend 1.2.0 then appends a closer after a glob like `` `foo/*` `` or `` `_meta` `` / `_blank`.

Finished markdown now stays `mode="static"`. Only the live streaming turn still remends. Reasoning and file previews are static too.

Any agent that writes those shapes could hit this. Grok just does it a lot.

Also fixes #555. That report looked like a separate code-fence parser bug: a reply ending exactly at a closing fence rendered ``` ```_ ``` as a line inside the block. Same cause. The appended `_` lands after the final fence, the fence stops closing, and the backticks become code content. Added a regression test for that shape (`message-final-fence.test.tsx`).

Measured truth table on the exact reply from #555, driving `MessageResponse` under vitest:

| mode | parseIncompleteMarkdown | result |
|---|---|---|
| streaming | true | ``` ```_ ``` inside the block |
| streaming | false | clean |
| static | true | clean |
| static | false (this PR) | clean |

Both defaults have to be at the bad value to reproduce, so either half of this PR is enough; it ships both.

Independent of the other open PRs. One concern.


xintaofei:
codeg work task `55` is done — https://github.com/xintaofei/codeg/pull/487 (4 files, +154/-20).

PR #490, opened 2026-08-16
feat(quota): show Cursor remaining from official DashboardService usage

Adam-Dalloul:
## Merge order

1. `#472` then `#481`
2. **Then this PR**

This branch already contains both. Merge those first so this reviews as the Cursor delta.

---

Cursor has no `usage` CLI. `status` and `about --format json` are identity only (this box is signed out, so I could not live-read a remaining number).

The official CLI/IDE client already ships `GetCurrentPeriodUsage` on `api2.cursor.sh`. Same class as the Claude `/usage` HUD: we only show remaining when that `planUsage` object is present (`total_percent_used`, or `remaining`/`limit`). Identity-only about/status payloads stay unavailable and the chip links the Spending dashboard.

Auth is the cursor-agent login file (`%APPDATA%\Cursor\auth.json` on Windows) or `CURSOR_API_KEY`. We do not read the IDE sqlite store.

This also maps the Cursor family onto the header chip from `#481`.


Adam-Dalloul:
Checked this against a signed-in Cursor CLI on this machine.

`cursor-agent status --format json` is authenticated. `%APPDATA%\Cursor\auth.json` has `accessToken`. `POST aiserver.v1.DashboardService/GetCurrentPeriodUsage` with that Bearer token returns 200 and `planUsage` (`remaining` / `limit` 40000, `totalPercentUsed` 0 on Ultra).

Pushed a follow-up that pins that live envelope and reads `billingCycleEnd` (unix ms string) as the reset time. Still no invented numbers, and we still do not read the IDE sqlite store.

PR #498, opened 2026-08-16
feat(cursor): show reasoning and context as separate chat options

Adam-Dalloul:
Settings -> Agents -> Cursor lists the full CLI catalog (`cursor-agent models`), including High / Fast / 1M / thinking variants. A new Cursor chat only gets the short ACP compatibility list, so the same model family looks incomplete.

Cursor documents the fix: advertise `_meta.parameterizedModelPicker: true` on initialize. Then ACP sends the base model plus separate options for reasoning, effort, thinking, context, and fast. The composer already renders every advertised session config option, so new Cursor variants show up without another host change.

This is the native hierarchical picker (pick the model, then optional effort / context). It is not the flattened CLI-row catalog in #489.

### DIFF
--- a/src-tauri/src/acp/connection.rs
+++ b/src-tauri/src/acp/connection.rs
@@ -3067,6 +3067,13 @@ fn claude_raw_sdk_session_meta(
 ///   Codex to keep the blast radius off other agents (e.g. Claude's native
 ///   AskUserQuestion, which would otherwise un-gate and duplicate the
 ///   codeg-mcp ask tool).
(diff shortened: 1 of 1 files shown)


Adam-Dalloul:
Closing: c2d72d31 already landed this, and more broadly. Main gates on uses_cursor_acp_backend rather than the built-in Cursor type, so custom agents wrapping cursor-agent acp get the picker too, and bddf6656 and ce08328e build on it. Nothing left here.

PR #499, opened 2026-08-16
Per-call permission_mode for delegated sub-agents

Adam-Dalloul:
## Problem

A parent agent can already choose *which* agent runs a delegated task, but not how much that child may do unattended. The child inherits whatever session mode the per-agent delegation default in Settings happens to be, which is the wrong granularity for a one-off delegation: you may want one specific child kept on a prompting or approval mode without changing the global default for that agent type.

This shows up when a parent fans work out to several children. Some slices are read-only research, others touch the filesystem, and today they all start in the same configured mode.

## Change

Adds an optional `permission_mode` to `delegate_to_agent`.

It is the target agent's own session mode id, the same vocabulary `AgentDelegationDefaults::mode_id` already uses, and it is forwarded verbatim as `ConnectionSpawner::spawn`'s existing `preferred_mode_id`. No new mechanism is introduced. This lets a caller reach the knob that `Settings > Delegation > per-agent defaults` already sets, but scoped to a single call.

| caller passes | result |
|---|---|
| omitted | configured Settings default, unchanged |
| a mode id | overrides the default for that one call |
| `""` or whitespace | treated as omitted |

Deliberate properties:

- **No behaviour change for existing callers.** Non-delegated sessions are untouched.
- **Blank is treated as omitted**, so a model emitting `""` cannot silently clear a configured default.
- **Agent agnostic.** The value is passed through rather than mapped through a hardcoded per-agent table, so it works for any agent that advertises session modes, and agents with none simply ignore it.
- The schema description states plainly that this is a cooperative permission scope enforced by the agent, not an OS sandbox, so it does not overclaim isolation.

## Tests

Three unit tests alongside the existing `agent_defaults_are_forwarded_to_spawner`:

- per-call value overrides the configured agent default
- omitting it keeps the configured default
- per-call value works when no agent default is configured

## Verification note

I could not run `cargo test` locally, as there is no Rust toolchain on this machine, so the Rust changes are unverified by compilation on my side and rely on CI. Everything else was checked by hand: the JSON schema parses, the schema diff is 4 added lines with formatting preserved, and every `DelegationRequest` construction site was audited for the new field. Flagging this explicitly rather than implying a green local run.

## Possible follow-up

Surfacing the mode a child actually started in on the delegation card would pair well with this, so a reviewer of a transcript can see what a delegated child was allowed to do. Kept out to keep one concern per PR. The per-agent default UI already exists in `delegation-agent-defaults.tsx`, so nothing is needed there.

### DIFF
--- a/src-tauri/src/acp/delegation/broker.rs
+++ b/src-tauri/src/acp/delegation/broker.rs
@@ -2296,11 +2296,15 @@ impl DelegationBroker {
         // Pull per-agent overrides from the broker config (defaults to empty).
         // Cloning is cheap — `AgentDelegationDefaults` is at most one Option<String>
         // and a small BTreeMap, and the spawner consumes both fields by value.
-        let (preferred_mode_id, preferred_config_values) = cfg
+        let (configured_mode_id, preferred_config_values) = cfg
             .agent_defaults
             .get(&req.agent_type)
             .map(|d: &AgentDelegationDefaults| (d.mode_id.clone(), d.config_values.clone()))
             .unwrap_or((None, BTreeMap::new()));
+        // A per-call `permission_mode` wins over the settings default; when the
+        // LLM omits it the configured default is used unchanged, so existing
+        // callers and non-delegated sessions see no behaviour change.
+        let preferred_mode_id = req.permission_mode.clone().or(configured_mode_id);
         // Checkpoint #1 (opportunistic): if a parent cancel already landed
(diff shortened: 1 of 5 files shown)


Adam-Dalloul:
Read side of this is #501, which shows the pinned mode on the delegation card. Merge this one first; #501 branches off it, so once this lands #501's diff collapses to just its own commit.

PR #501, opened 2026-08-16
Show the pinned session mode on the delegation card

Adam-Dalloul:
Stacks on #499, which adds the optional `permission_mode` argument. This PR is the read side of it.

## Problem

When a parent pins a session mode for one delegated child, nothing in the transcript records it. Reviewing a conversation afterwards you can see which agent ran and what it was asked to do, but not what it was allowed to do without being asked. That is the one detail you want when auditing what a sub-agent was permitted to touch.

## Change

A quiet monospace chip beside the task id on the delegation card, rendered only when the parent actually pinned a mode.

- No chip means the child used the configured per-agent default, which is the common case, so existing transcripts look unchanged.
- The value comes from the already-parsed `permission_mode` argument, so there is no new wire field, no broker change, and no migration.
- Hosts that strip tool arguments simply show no chip, which reads correctly rather than misleading.
- `parseInput` now also recognises a lone `permission_mode` as delegation args, matching how `task` / `agent_type` / `working_dir` are each treated as sufficient.

The tooltip string is added to all ten locale files with the English text, for translators to pick up.

## Tests

Three added to `delegation-card.test.ts`: parses and trims the value, treats a blank string as absent, leaves it null when the argument is omitted.

## Verification

Run locally on this branch:

- `npx tsc --noEmit -p tsconfig.json` clean, exit 0
- `npx vitest run src/components/message/ src/lib/delegation-card.test.ts` 31 files, 344 tests, all passing

`npx eslint` reports `Delete ␍` on these files, but it does so on untouched files too (276 on `delegation-settings.tsx`), so that is this Windows checkout's CRLF state rather than anything in this change. The committed diff is 57 insertions with no line-ending churn.

### DIFF
--- a/src-tauri/src/acp/delegation/broker.rs
+++ b/src-tauri/src/acp/delegation/broker.rs
@@ -2463,11 +2463,15 @@ impl DelegationBroker {
         // Pull per-agent overrides from the broker config (defaults to empty).
         // Cloning is cheap — `AgentDelegationDefaults` is at most one Option<String>
         // and a small BTreeMap, and the spawner consumes both fields by value.
-        let (preferred_mode_id, preferred_config_values) = cfg
+        let (configured_mode_id, preferred_config_values) = cfg
             .agent_defaults
             .get(&req.agent_type)
             .map(|d: &AgentDelegationDefaults| (d.mode_id.clone(), d.config_values.clone()))
             .unwrap_or((None, BTreeMap::new()));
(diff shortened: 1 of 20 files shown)


Adam-Dalloul:
Review order: merge #499 first. Cross-fork PRs cannot use a fork branch as their base, so this is opened against main and currently shows #499's commit too. Once #499 lands, this diff collapses to its own single commit. Only the second commit, 'show the pinned session mode on the delegation card', is new here.

PR #502, opened 2026-08-16
feat(chat): jump to a tab with Ctrl/Cmd+1-9 and reopen with Ctrl/Cmd+Shift+T

Adam-Dalloul:
There was no numbered tab jump, only Ctrl/Cmd+Tab and Shift+Tab to cycle. There was also no way to get back a tab you just closed.

This adds the usual browser tab keys in the same always-mounted workspace controller as close/next/prev, so they still work when the tab strip is hidden.

- Ctrl/Cmd+1 through 9 jump to that conversation tab, or file tab when the file pane is active. Missing tab is a no-op.
- Ctrl/Cmd+Shift+T reopens the most recently closed conversation or file tab. Repeat to walk back through the last 20. A dirty file tab that you declined to close is not recorded.

Both bindings show up in Settings so they can be remapped.


Adam-Dalloul:
Rebuilt on current main. The branch had drifted a long way behind, so this is now just its two commits cherry-picked onto main rather than a replay of everything in between.

One thing worth your call before this lands. Your note on matchesDigitRowCode says the unshifted half of the positional fallback is latent because nothing binds digits 1-9. This PR binds all nine, so it stops being latent: on AZERTY, - sits on Digit6, so Ctrl+- matches mod+- by character and mod+6 by position, and one press would zoom out and jump to tab 6. Same shape for _ on Digit8. shortcutsConflict still cannot see it, for the reason you gave.

I have not tried to fix it here since it is your design call which side should win. Happy to gate the positional fallback behind no other action matching by character, or to drop the fallback for digits other than 0, whichever you prefer.

xintaofei:
codeg work task `123` is done — https://github.com/xintaofei/codeg/pull/502 (13 files, +465/-24).

PR #504, opened 2026-08-16
fix(chat): name image cards after the tool or page, not Image generation

Adam-Dalloul:
The in-position image card always printed Image generation. That is right for Codex image generation. It is wrong for a Read of a screenshot or a fetched page.

The card still shows the picture in place. The heading now uses the tool title when it is not that hardcoded Codex string, or the URL slug / filename from the tool input. Real generation is unchanged.


xintaofei:
codeg work task `120` is done — https://github.com/xintaofei/codeg/pull/504 (4 files, +254/-20).

PR #511, opened 2026-08-18
Drop a stranded background-tasks chip after keepalive

Adam-Dalloul:
The sky bar should leave the moment the work is actually done. Same rule for every agent that can raise it.

What raises the bar today: Claude background shells / async sub-agents, and Grok background commands (`task_backgrounded`) plus background `spawn_subagent` kids. Codex, Cursor, Gemini, and OpenCode never raise it.

What now drops it immediately:
- Claude: BashOutput with an exit code, TaskOutput terminal / success, TaskStop / KillShell, or a task-notification
- Grok: `task_completed` (the process exited) or `subagent_finished`

The one-hour keepalive is only the last resort if a done signal never arrives. A snapshot will not bring a dead count back on reconnect.

### DIFF
--- a/src-tauri/src/acp/background_watch.rs
+++ b/src-tauri/src/acp/background_watch.rs
@@ -1312,6 +1312,11 @@ fn is_terminal_task_status(status: &str) -> bool {
     matches!(
         status,
         "completed"
+            | "complete"
+            | "success"
+            | "succeeded"
+            | "done"
+            | "exited"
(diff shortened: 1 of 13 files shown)


Adam-Dalloul:
Closing this one: main removed the background-tasks strip entirely in cb8ab399, which resolves the stranded-count problem this was fixing. Nothing left to rebase onto.

PR #530, opened 2026-08-21
Don't count tool waits in output speed

Adam-Dalloul:
The live tok/s number was folding in time spent waiting on tools, so a fast model looked slow whenever a command was running.

This pauses the clock while a tool is in flight, holds the last generating rate, and keeps a session average of tokens over generating time only. That average shows up in Session Details, the composer token popover, and the bottom-left status bar.

A slow stream with no tool still counts the quiet gaps, so a trickle is not reported as fast.


Adam-Dalloul:
Pulled it off the status bar. Average tok/s now lives in Session Details and on the Token Usage page.

PR #531, opened 2026-08-21
Polish model option labels so the trigger matches the row

Adam-Dalloul:
Claude's composer list ships a short name and puts the version in the description, so the chip says Fable while the row says Fable 5. Default's blurb is also just the next row's title, which makes the list look like two copies of the same model.

Display only. Values stay the same. Same polish on the wide chips, the searchable picker, and the collapsed rail so they cannot drift.

This is not the Cursor Settings vs composer catalog gap in #498. That one still needs the parameterized picker. This just cleans the labels every agent already sends.

### DIFF
--- a/src/components/chat/message-input.tsx
+++ b/src/components/chat/message-input.tsx
@@ -86,7 +86,6 @@ import {
   type SessionSelectorSetting,
 } from "@/components/chat/session-selectors-panel"
 import {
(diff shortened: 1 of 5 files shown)


Adam-Dalloul:
Closing this. Rewriting labels on our side was the wrong idea. The composer should show whatever the agent actually advertises.

PR #540, opened 2026-08-21
feat(settings): let the sidebar hide Automations, To-dos, and Repository panel

Adam-Dalloul:
The sidebar's fixed rows always render, and there was no way to hide the ones you never use. This adds a Sidebar navigation section to Settings > Appearance with one switch each for Automations, To-dos, and Repository panel.

Defaults are unchanged: every row stays visible until you turn it off. New chat is deliberately not hideable, and hidden pages stay reachable from the status bar's quick actions menu. The preference is a single localStorage record synced across windows like the other appearance settings, so hiding a row in the settings window updates the workspace immediately. Hidden rows unmount rather than hide with CSS.

Strings are added to all ten locales. Tests cover the parse fallbacks, the storage round-trip with cross-window sync, and the sidebar gating; the full vitest suite passes (4705 tests).


Adam-Dalloul:
Closing this one: main now has these toggles in the sidebar view-options menu, driven by SIDEBAR_NAV_ITEM_IDS so a new route cannot ship without one. That covers what this PR was for, and the id-driven list is the better shape.

PR #556, opened 2026-08-24
Right-click Copy image on staged composer attachments

Adam-Dalloul:
## Symptom

Right-clicking an image that is STAGED in the composer (a pasted or attached picture, or its blown-up preview) opens the composer's text context menu (Cut / Copy / Paste as plain text / Select all / Quick messages). The Copy row copies nothing for an image, and there is no Copy image row at all.

## Cause

#482 wired the Copy/Download image actions onto the transcript surfaces (generated images, user attachments in messages, and their preview dialogs). The composer's own attachment strip (`ComposerImageThumbnails`) and the preview it opens were never covered, so a right-click there bubbles up to `message-input`'s text menu.

## Fix

Wrap each staged tile in the same `ImageActions` context menu, and pass the preview dialog the same `onCopy` / `onDownload` / `renderImage` it already supports. The menu reuses the transcript's translation keys, so no new locale strings and both menus stay identical in every language. The non-secure-context fallback (native browser image menu instead of a Copy row that can only fail) carries over for free.

## Validation

- 3 new tests on the strip: right-click offers Copy image and Download image without triggering the composer's ancestor menu, copy/download hit the shared clipboard/download helpers with the staged bytes, and the no-clipboard passthrough still blocks the text menu.
- Full `src/components/message` + `src/components/chat/composer` suites: 715 tests pass.
- `eslint` and `tsc --noEmit` clean.

### DIFF
--- a/src/components/chat/composer/composer-image-thumbnails.test.tsx
+++ b/src/components/chat/composer/composer-image-thumbnails.test.tsx
@@ -1,0 +1,118 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react"
+import { NextIntlClientProvider } from "next-intl"
+import { beforeEach, describe, expect, it, vi } from "vitest"
+
+import enMessages from "@/i18n/messages/en.json"
+import type { ImageInputAttachment } from "@/components/chat/message-input-attachments"
+
+const mocks = vi.hoisted(() => ({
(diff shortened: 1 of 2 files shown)


xintaofei:
codeg work task `94` is done — https://github.com/xintaofei/codeg/pull/556 (2 files, +213/-2).

PR #565, opened 2026-08-24
fix(window): show a hidden owner when its auxiliary window closes

Adam-Dalloul:
Hit this on Windows 11: open Settings, then close the workspace window (its X hides to tray, which is easy to do by accident with Settings sitting on top of it). Close Settings and the app is now on screen nowhere.

`restore_windows_after_settings` only calls `set_focus()` on the owner window. That call is skipped outright when the window is hidden or minimized, on Windows and on macOS both, so a workspace that got hidden to the tray never comes back and you are left with no visible window. Settings is intentionally not a child window, so nothing else pulls it back either.

`show_main_window` in the same file already has the right sequence for the tray and dock paths, so this pulls that into a small helper and uses it for the settings, commit and merge owner restores, which all had the same bare `set_focus`.

Two unit tests for the owner map, since the restore now has a visible effect and the handler fires on both CloseRequested and Destroyed.

### DIFF
--- a/src-tauri/src/commands/windows.rs
+++ b/src-tauri/src/commands/windows.rs
@@ -826,18 +826,25 @@ pub async fn open_import_sessions_window(
     Ok(())
 }
 
-/// Bring an auxiliary window's owner back to the front when that auxiliary
-/// window closes.
+/// Bring `label` to the foreground: unminimize, unhide, then focus.
 ///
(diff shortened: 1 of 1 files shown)


xintaofei:
codeg work task `101` is done — https://github.com/xintaofei/codeg/pull/565 (1 file, +59/-14).

PR #566, opened 2026-08-24
fix(acp): adopt a Claude title written after the turn ends

Adam-Dalloul:
Claude Code has no ACP event for its generated title. Its adapter reads the name back out of the session file at turn-end, and the SDK writes that name in the background, so on a short chat the turn is over before there is anything to read. Nothing publishes the title and the conversation keeps its first-prompt name until the next time you open it.

The transcript watcher is already tailing those exact bytes for background activity. It now also folds `ai-title` / `custom-title` records into the session's name and hands the result to the same path a live ACP title takes, so the sidebar updates a beat after Claude generates the name instead of on the next detail load.

Resolution is `customTitle ?? aiTitle`, matching what the parser already does over the whole file, so a `/rename` still outranks a generated title. Blank values are ignored, an unchanged name is not republished, and history before the watch baseline is left to the ordinary detail fetch. The write itself is unchanged: `refresh_auto_title` still skips a conversation the user renamed and still no-ops when the value has not moved.

The skip-cache and the unbound-row drop moved into `session_title::publish_native_title` so the ACP path and this one cannot drift apart.

### DIFF
--- a/src-tauri/src/acp/background_watch.rs
+++ b/src-tauri/src/acp/background_watch.rs
@@ -37,13 +37,23 @@
 //!   consumed exactly once, so a cron//loop re-fire of the SAME text later
 //!   correctly classifies as out-of-turn).
 //!
+//! * **Session title** — Claude Code's generated name arrives as a dedicated
+//!   `ai-title` transcript record whenever the background summarizer finishes,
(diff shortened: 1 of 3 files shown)


xintaofei:
codeg work task `100` is done — https://github.com/xintaofei/codeg/pull/566 (3 files, +568/-26).

PR #569, opened 2026-08-24
feat(settings): raise window zoom to 300%

Adam-Dalloul:
Follow-up to #484, split out as you asked.

Raises the top window zoom rung to 300%. Same discrete rungs and the same Ctrl/Cmd + and - stepping, just more headroom at the high end for large displays.

Stacks on #484, so it is easiest to take after that one lands.


xintaofei:
Thanks for splitting this out — the scope is exactly right. Three files, one concern, and the change reads in about twenty seconds.

Verified:

- **`VALID_ZOOMS` and `ZOOM_LEVELS` are byte-identical** (`[80,90,100,110,125,150,175,200,250,300]`). That's the one I most wanted to check, since the pre-paint inline script is what actually applies the zoom on boot.
- **Stepping covers the new range**: 7 presses from 100% to 300%, 9 back down to 80%, clamping correctly at both ends.
- The Rust side already clamps to 50–300, so the traffic-light sync needs nothing.
- Monaco and xterm both scale their font size off `zoomLevel`, so the new high rungs don't leave the editor and terminal behind.
- Full gate green: 353 files / 4865 tests, `tsc --noEmit` clean, ESLint clean, Prettier clean.

Worth noting the two PRs turned out to complement each other nicely: 300% is a level you can genuinely paint yourself into a corner with on a small window, and ⌘0 from #484 is the way back out.

## Two notes, neither blocking

**Stacking.** This branch is based on `2d2f2a1b`, so it doesn't include #484's newest commit (`1bc10ff8`, the digit-row fix). I simulated both landing orders:

- **Merge commit** (what this repo uses): merging #484 then #569 is clean — no conflicts, the digit-row fix survives, 300% lands correctly. So nothing to do.
- **Squash-merge #484 first**: conflicts in `keyboard-shortcuts.ts`, `keyboard-shortcuts.test.ts` and `theme-presets.test.ts`.

So it works as-is on our normal path; just flagging it in case anyone reaches for Squash in the UI, in which case a rebase onto main first is the easy fix.

**A parity test would be cheap insurance.** Nothing currently guards `VALID_ZOOMS` against `ZOOM_LEVELS`, and this PR is precisely the kind of change that touches one and could forget the other. The failure mode is quiet rather than loud: the provider derives its state from the font size the inline script already set (`appearance-provider.tsx`, the `useState<ZoomLevel>` initializer), so an unknown stored level doesn't flash — it just renders at 100% and the user's saved zoom is silently ignored. `APPEARANCE_INIT_SCRIPT` is exported, so it's about five lines:

```ts
it("keeps the pre-paint whitelist in step with ZOOM_LEVELS", () => {
  const listed = /var VALID_ZOOMS = \[([^\]]*)\]/
    .exec(APPEARANCE_INIT_SCRIPT)![1]
    .split(",")
    .map((n) => parseInt(n.trim(), 10))
  expect(listed).toEqual([...ZOOM_LEVELS])
})
```

Entirely up to you whether it goes in here or not — happy to merge without it.

Thanks again for doing the split; it made both of these much easier to reason about.

Adam-Dalloul:
Added the parity test, and confirmed it actually bites by dropping 300 from VALID_ZOOMS and watching it fail. Worth having given the failure is silent rather than loud.

Also rebased onto main now that #484 has landed, so the stacking caveat is gone and this is a single commit on top of current main plus the test. Full suite 353 files / 4868 green, tsc and Prettier clean.

xintaofei:
Perfect — and thank you for going the extra step of confirming the test actually bites rather than just that it passes. A green assertion that would stay green through the regression it's named after is the most common way this kind of guard turns into decoration, and you checked. That's the right instinct.

Verified on `7c02d45d` before landing:

- Rebase is clean: three files, +24/-4 against current main, single concern.
- `VALID_ZOOMS` and `ZOOM_LEVELS` byte-identical at all ten rungs.
- Reproduced the guard failing on drift myself — dropping just `300` from the script's list fails with a clear diff, so it's genuinely load-bearing.
- Full suite 353 files / 4868 tests, `tsc --noEmit` clean, ESLint clean, Prettier clean.
- All seven CI checks green.

Merging. Thanks for seeing this one all the way through — five rounds on #484 and a clean split here, each round coming back with the fix, a test that pins it, and an honest note about what the fix doesn't cover. That last part is rarer than it should be and it's what made this straightforward to land.

PR #577, opened 2026-08-25
feat(remote-workspace): add custom HTTP headers to connections

codebutler:
## Why

A user can put a reverse proxy in front of a codeg-server. The proxy can require extra HTTP headers, for example a Cloudflare Access service token. Codeg could not send those headers, so the connection failed.

## What

Each remote workspace connection can now carry custom HTTP headers. The desktop client attaches them to every request it sends to that connection.

The manage dialog gets a "Custom headers" section below the access token. The section stays collapsed unless the connection already has headers. Each row holds a name and a value. Values are masked, because a header usually carries a credential.

image

## How

- A new migration adds a `headers` column. It stores a JSON array of `{"name","value"}` objects in a TEXT column, which follows the convention of the other JSON blobs in this schema. Existing rows default to `[]`.
- An array, not a map: HTTP allows a repeated header name, and the array keeps the order the user typed.
- `RemoteWorkspaceHeader` lives in `models/`. It converts to an `http` pair through `to_header_pair`, and a slice of rows converts to a `HeaderMap` through the `ToHeaderMap` trait.
- The service validates each row before it writes: it drops the blank rows the editor leaves, rejects a value with no name, rejects a name or value that `http` cannot parse, and rejects 10 reserved names. The client sets those names itself, so a custom one would break the request. The save fails rather than silently drop what the user typed.
- `commands/remote_proxy.rs` attaches the headers at all five request sites: the HTTP call, the attachment upload, the workspace file upload, the download ticket, and the ticket download. The last one sends no bearer token, because the ticket in the URL is the credential — but a fronting proxy still inspects it, so it needs the headers too.
- The WebSocket handshake carries the same headers.
- The health check sends the headers as well, so the test and the save exercise the same configuration as a later request.

## Tests

Rust, in `remote_workspace_connection_service`:

- `validate_headers_trims_and_drops_empty_rows`
- `validate_headers_keeps_repeated_names_and_order`
- `validate_headers_rejects_reserved_names`
- `validate_headers_rejects_invalid_name_value_and_missing_name`
- `to_header_map_keeps_repeats_and_skips_unparsable_rows`
- `create_list_update_delete_roundtrip`, extended to store and read back the headers

Frontend, in the new `remote-workspace-manage-dialog.test.tsx`:

- `keeps the editor collapsed when the connection has no headers`
- `opens the editor when the connection already has headers`
- `masks every header value`
- `sends the added header on save and drops a removed one`

### DIFF
--- a/src-tauri/src/commands/remote_proxy.rs
+++ b/src-tauri/src/commands/remote_proxy.rs
@@ -141,6 +141,78 @@ struct WsSubscriber {
     window_instance_id: String,
 }
 
+/// Matches `reqwest`'s own default redirect bound.
+const MAX_REMOTE_REDIRECTS: usize = 10;
+
+/// Two URLs reach the same server. Host and port only, deliberately: this is
+/// the exact test `reqwest` applies when it decides whether to strip
+/// `Authorization` across a redirect, so an `http` → `https` hop reads as a
+/// different origin through its default port, the same way it does there.
+fn same_remote_origin(a: &reqwest::Url, b: &reqwest::Url) -> bool {
+    a.host_str() == b.host_str() && a.port_or_known_default() == b.port_or_known_default()
+}
+
+/// `host:port` for an error message. The port is not decoration: two ports on
+/// one host are two origins here, and "refused a redirect off 127.0.0.1 to
+/// 127.0.0.1" would read as a bug.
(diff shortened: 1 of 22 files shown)


xintaofei:
codeg work task `105` is done — https://github.com/xintaofei/codeg/pull/577 (22 files, +1168/-30).

PR #584, opened 2026-08-26
feat(experts): add agent rules picker

n-shamsi:


n-shamsi:
add rules picker

n-shamsi:
Hmm, I don't think it makes sense to introduce this under Experts for the majority of users. Closing this request.

PR #587, opened 2026-08-27
fix(acp): preserve mode across fork

### LINKED ISSUE #398: Forked Claude Code conversation always starts with "Manual" mode
## Problem
Claude Code 对话使用 "Fork & Send" 创建的新对话，权限模式总为 "Manual"

## Expected behavior
继承原对话的权限模式

dawNotPoi:
## Summary

- capture the live parent ACP mode when a fork is requested
- reapply a differing mode before selectors become ready and queued prompts run
- fail closed without releasing queued prompts if the inherited mode cannot be restored
- ignore stale ancestor state when the current parent does not advertise modes
- add regression coverage for differing, matching, absent, stale, latest, ordered, and rejected mode restoration

## Why

A fork response can reset Claude Code to its default Manual mode while the frontend still caches the parent selection. The next prompt can therefore run under the wrong permission mode. Preserving the backend live mode across the fork keeps the child session behavior consistent with its parent. If the agent rejects that restoration, the queued prompt is blocked and the connection reports a terminal error instead of silently executing under the default mode.

## Testing

- `cargo test --no-default-features --lib fork_mode_inheritance` (8 passed)
- `cargo test --no-default-features --lib` (3248 passed, 1 ignored)
- `cargo test --no-default-features --bin codeg-server`
- `git diff --check`

Fixes #398


xintaofei:
The "Fork & Send" button has been removed from the conversation input box; however, "forking" is now supported within the conversation messages themselves, without the previously mentioned issues.

Thanks to @dawNotPoi for the contribution, though it is no longer applicable.

PR #593, opened 2026-08-28
fix(codex): skip native subagent session imports

TroyMitchell911:
## Summary

- preserve Codex parent thread metadata from both session_meta shapes
- expose it in list and detail summaries so the generic importer filters child rollouts
- share the same normalized detection with forked-thread statistics

## Why

Codeg scans native Codex rollout files directly instead of going through the ACP session list. The parser previously hardcoded parent_id to null, so native subagent rollouts appeared as independent root sessions and could be imported incorrectly.

## Tests

- 2 parent-thread parser regressions
- 19 delegation-child/import regressions
- full Rust library suite: 3090 passed, 1 ignored
- server and desktop clippy with warnings denied
- desktop cargo check and production frontend build

### DIFF
--- a/src-tauri/src/parsers/codex.rs
+++ b/src-tauri/src/parsers/codex.rs
@@ -88,6 +88,8 @@ impl CodexParser {
 
         let mut conversation_id: Option<String> = None;
         let mut cwd: Option<String> = None;
+        let mut parent_id: Option<String> = None;
+        let mut session_header_seen = false;
         let mut git_branch: Option<String> = None;
         let mut model: Option<String> = None;
         let mut title: Option<String> = None;
@@ -165,6 +167,16 @@ impl CodexParser {
                             .get("cwd")
                             .and_then(|s| s.as_str())
                             .map(|s| s.to_string());
+                        // Only the FIRST header names this thread's parent. A
+                        // forked child replays the parent's header further down
+                        // the same file and that copy declares no parent, so a
+                        // last-one-wins read would clear the child's `parent_id`
+                        // and hand it back to the importer as a root session.
+                        // Same rule `parse_codex_subagent_stats` uses.
+                        if !session_header_seen {
+                            session_header_seen = true;
+                            parent_id = codex_parent_thread_id(payload);
+                        }
                         _cli_version = payload
                             .get("cli_version")
                             .and_then(|s| s.as_str())
@@ -393,7 +405,7 @@ impl CodexParser {
             message_count,
             model,
             git_branch,
-            parent_id: None,
+            parent_id,
             parent_tool_use_id: None,
             delegation_call_id: None,
         }))
@@ -1868,10 +1880,40 @@ fn build_collab_wait_input(status: &serde_json::Map<String, serde_json::Value>) 
     (input.to_string(), any_error)
 }
 
-/// Whether a transcript's opening record declares it a forked thread
-/// (`session_meta.parent_thread_id`) — codex 0.147's sub-agent shape, where the
-/// child is a rollout of its own and `fork_turns` copies the parent's history
-/// into its head.
+/// The parent thread id a rollout's `session_meta` payload declares, or `None`
+/// for a root session.
+///
+/// TWO shapes carry it and both are live on disk, so neither branch may be
+/// dropped. The structured `source.subagent.thread_spawn.parent_thread_id` is
+/// the one every sub-agent rollout has (checked across on-disk rollouts from
+/// 0.117 through 0.147); the flat `parent_thread_id` is a later, additive
+/// mirror that only some versions write, always alongside the structured form
+/// and never on its own. Reading only the flat field — as this did before —
+/// therefore misses real sub-agents, which is what let their rollouts list as
+/// importable root sessions and let their replayed history be counted as their
+/// own (see [`is_forked_thread_header`]).
+///
+/// A present-but-blank id is treated as absent: it names no parent, and a
+/// whitespace `parent_id` would still read as "this is a child" downstream.
+fn codex_parent_thread_id(payload: &serde_json::Value) -> Option<String> {
+    payload
+        .get("parent_thread_id")
+        .and_then(serde_json::Value::as_str)
+        .map(str::trim)
+        .filter(|id| !id.is_empty())
+        .or_else(|| {
+            payload
+                .pointer("/source/subagent/thread_spawn/parent_thread_id")
+                .and_then(serde_json::Value::as_str)
+                .map(str::trim)
+                .filter(|id| !id.is_empty())
+        })
+        .map(ToOwned::to_owned)
+}
+
+/// Whether a transcript's opening record declares it a FORKED thread — codex
+/// 0.147's `fork_turns` sub-agent shape, where the child is a rollout of its own
+/// and the parent's history is copied into its head.
 ///
 /// That copy is the problem: the parent's own tool calls sit at the top of the
 /// child's file, indistinguishable at the record level from the child's, so
@@ -1884,11 +1926,27 @@ fn build_collab_wait_input(status: &serde_json::Map<String, serde_json::Value>) 
 /// still names the sub-agent and badges its thread id (both come from the
 /// PARENT's rollout); only the nested tool-call list is absent. The legacy
 /// `agent-<id>.jsonl` shape has no such prefix and is unaffected.
+///
+/// BOTH markers are required, because "is a sub-agent" and "replays the parent"
+/// are different things and only the second one may be refused. A parent thread
+/// id alone says a thread was spawned by another; `forked_from_id` is codex's
+/// own declaration that this rollout was seeded with that thread's history. On
+/// disk they split cleanly — of 46 sub-agent rollouts, the 23 carrying
+/// `forked_from_id` are exactly the 23 that also replay a second `session_meta`
+/// header, and the other 23 are ordinary children whose tool calls are their
+/// own. Refusing on the parent id alone silently blanks the capsule for those.
 fn is_forked_thread_header(value: &serde_json::Value) -> bool {
-    value.get("type").and_then(|t| t.as_str()) == Some("session_meta")
-        && value
-            .pointer("/payload/parent_thread_id")
-            .is_some_and(|v| v.as_str().is_some_and(|s| !s.is_empty()))
+    if value.get("type").and_then(|t| t.as_str()) != Some("session_meta") {
+        return false;
+    }
+    let Some(payload) = value.get("payload") else {
+        return false;
+    };
+    codex_parent_thread_id(payload).is_some()
+        && payload
+            .get("forked_from_id")
+            .and_then(serde_json::Value::as_str)
+            .is_some_and(|id| !id.trim().is_empty())
 }
 
 fn parse_codex_subagent_stats(
@@ -2147,6 +2205,8 @@ impl CodexParser {
 
         let mut messages = Vec::new();
         let mut cwd: Option<String> = None;
+        let mut parent_id: Option<String> = None;
+        let mut session_header_seen = false;
         let mut git_branch: Option<String> = None;
         let mut model: Option<String> = None;
         let mut title: Option<String> = None;
@@ -2346,6 +2406,13 @@ impl CodexParser {
                             .get("cwd")
                             .and_then(|s| s.as_str())
                             .map(|s| s.to_string());
+                        // First header wins — see the same latch in
+                        // `parse_jsonl_summary`: the replayed parent header that
+                        // follows a forked child's own declares no parent.
+                        if !session_header_seen {
+                            session_header_seen = true;
+                            parent_id = codex_parent_thread_id(payload);
+                        }
                         git_branch = payload
                             .get("git")
                             .and_then(|g| g.get("branch"))
@@ -3823,7 +3890,7 @@ impl CodexParser {
             message_count: turns.len() as u32,
             model,
             git_branch,
-            parent_id: None,
+            parent_id,
             parent_tool_use_id: None,
             delegation_call_id: None,
         };
@@ -4817,6 +4884,7 @@ mod tests {
     use super::extract_context_window_used_tokens_from_token_count_info;
     use super::extract_response_item_user_image_blocks;
     use super::extract_turn_usage_from_codex_usage;
+    use super::codex_parent_thread_id;
     use super::is_encrypted_envelope;
     use super::merge_codex_context_window_stats;
     use super::native_team_wait_input;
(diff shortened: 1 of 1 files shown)


xintaofei:
Thanks for this — the diagnosis is right and the vehicle is the right one. `parent_id` is exactly how `opencode.rs` and `hermes.rs` mark a non-root session, and `import_service::collect_local_summaries` already drops `parent_id.is_some()`, so nothing new had to be invented.

I reviewed it against the rollouts on my machine rather than only reading the diff — 2,899 files under `~/.codex/sessions`, cli 0.94 → 0.149, 46 of them sub-agent rollouts. Two things came out of that which I think are worth folding in before this lands, plus one nice surprise in your favour.

I did try to push the follow-up straight to `fix/codex-subagent-import`, but the fork has *Allow edits by maintainers* turned off:

```
git push failed — pushing to a fork needs its author to allow edits from
maintainers on the pull request
```

So the patch is at the bottom of this comment instead. If you tick that box on the PR sidebar I'm happy to push it for you; otherwise `git apply` should be clean — I checked it against `536600ce`, and the rebase didn't touch `codex.rs`, so it still applies as-is.

---

### 1. The replayed parent header clears `parent_id` again

This is the one that matters most, because it lands on exactly the files the PR is aimed at.

`fork_turns` copies the parent's history into the child's head, and that copy includes **the parent's own `session_meta` record**. That second header declares no parent of its own. Both summary parsers assign on *every* `session_meta`:

```rust
"session_meta" => {
    if let Some(payload) = value.get("payload") {
        conversation_id = ...;
        cwd = ...;
        parent_id = codex_parent_thread_id(payload);   // last header wins
```

so the child's `parent_id` gets set by its own header and then wiped by the replayed one. 23 of the 46 sub-agent rollouts here carry that second header. Measured by pointing `CodexParser::with_base_dir` at the real corpus:

| | rollouts reporting `parent_id` |
|---|---|
| this PR as it stands | 23 / 46 |
| with a first-header latch | 46 / 46 |

The existing test passes because its fixture has a single header. The fix is to latch on the first one, which is the same rule `parse_codex_subagent_stats` already applies via `checked_header`:

```rust
if !session_header_seen {
    session_header_seen = true;
    parent_id = codex_parent_thread_id(payload);
}
```

### 2. `is_forked_thread_header` now also suppresses stats for clean children

Reusing the helper inside `is_forked_thread_header` widens it from "has a flat parent id" to "has any parent id" — but *being a sub-agent* and *replaying the parent's history* turn out to be different populations, and only the second may be refused.

The corpus splits perfectly, with no overlap either way:

- **23** rollouts carry `forked_from_id` — these are exactly the 23 that also hold the replayed second header. Their prefix really is the parent's work, so refusing them is correct.
- **23** carry a parent thread id and *no* `forked_from_id` and no replayed prefix. They're ordinary spawned children whose tool calls are their own — 81 calls across 21 files.

5 of that second group were already being refused before this PR (they happen to carry the flat mirror), so the net effect of the change is 18 more files silently dropped from the collab capsule's tool list — 76 tool calls across the 16 of them that have any.

Requiring both markers restores the split cleanly — verified by calling `parse_codex_subagent_stats` for each of the 46 real thread ids in its real directory: **23 replayed refused, 23 clean counted, no misclassification in either direction.**

```rust
codex_parent_thread_id(payload).is_some()
    && payload
        .get("forked_from_id")
        .and_then(serde_json::Value::as_str)
        .is_some_and(|id| !id.trim().is_empty())
```

### The nice surprise

Your fallback branch fixes more than the PR claims. The *old* flat-only `is_forked_thread_header` was silently under-firing on 22 real sub-agent rollouts — `parse_codex_subagent_stats` resolves a sub-agent's file by "stem ends with the thread id", which matches `rollout-<ts>-<thread_id>.jsonl`, and I confirmed all 22 sit in the same directory as their parent. They were being counted, crediting each child with the parent's replayed tool calls. Worth calling out in the description.

### Smaller notes

- The new `codex_parent_thread_id` landed *between* the old doc comment and `is_forked_thread_header`, so the "a forked thread yields no stats rather than wrong ones" rationale now documents the id accessor and `is_forked_thread_header` is undocumented. The patch moves it back and gives the helper its own.
- One wording thing: the new doc says codex "writes `session_meta.parent_thread_id` today and keeps the same id under the structured subagent source for compatibility." It's the other way round — the structured `source.subagent.thread_spawn` form is on all 46 and spans 0.117 → 0.147, while the flat field is on 24 and **never appears without** the structured one (the 24 pairs agree on the value). Reading it as written would invite someone to delete the branch that's actually carrying the feature.
- The end-to-end test only exercised the flat shape, i.e. the one shape that never occurs alone. The patch turns it into a two-case loop and adds a clean-child-keeps-its-stats case; the two replay fixtures also gain `forked_from_id`, which real replayed rollouts always have.

Verification on my side: 3122 desktop lib tests green, desktop + server clippy with `-D warnings`, server and `codeg-mcp` checks. I also confirmed both new assertions actually fail without the corresponding fix, so they're load-bearing rather than decorative.

---

<details>
<summary>Follow-up patch (applies onto <code>536600ce</code>)</summary>

```diff
diff --git a/src-tauri/src/parsers/codex.rs b/src-tauri/src/parsers/codex.rs
index a4b6d814..a0735c7b 100644
--- a/src-tauri/src/parsers/codex.rs
+++ b/src-tauri/src/parsers/codex.rs
@@ -89,6 +89,7 @@ impl CodexParser {
         let mut conversation_id: Option<String> = None;
         let mut cwd: Option<String> = None;
         let mut parent_id: Option<String> = None;
+        let mut session_header_seen = false;
         let mut git_branch: Option<String> = None;
         let mut model: Option<String> = None;
         let mut title: Option<String> = None;
@@ -166,7 +167,16 @@ impl CodexParser {
                             .get("cwd")
                             .and_then(|s| s.as_str())
                             .map(|s| s.to_string());
-                        parent_id = codex_parent_thread_id(payload);
+                        // Only the FIRST header names this thread's parent. A
+                        // forked child replays the parent's header further down
+                        // the same file and that copy declares no parent, so a
+                        // last-one-wins read would clear the child's `parent_id`
+                        // and hand it back to the importer as a root session.
+                        // Same rule `parse_codex_subagent_stats` uses.
+                        if !session_header_seen {
+                            session_header_seen = true;
+                            parent_id = codex_parent_thread_id(payload);
+                        }
                         _cli_version = payload
                             .get("cli_version")
                             .and_then(|s| s.as_str())
@@ -1870,23 +1880,21 @@ fn build_collab_wait_input(status: &serde_json::Map<String, serde_json::Value>)
     (input.to_string(), any_error)
 }
 
-/// Whether a transcript's opening record declares it a forked thread. Codex
-/// writes `session_meta.parent_thread_id` today and keeps the same id under the
-/// structured subagent source for compatibility. This is codex 0.147's
-/// sub-agent shape, where the child is a rollout of its own and `fork_turns`
-/// copies the parent's history into its head.
+/// The parent thread id a rollout's `session_meta` payload declares, or `None`
+/// for a root session.
 ///
-/// That copy is the problem: the parent's own tool calls sit at the top of the
-/// child's file, indistinguishable at the record level from the child's, so
-/// counting the whole file would credit the child with work the PARENT did. The
-/// hand-off that separates them (the addressed inter-agent `agent_message`) is
-/// not a reliable boundary either — a parent that already collected an earlier
-/// sibling's reply carries one inside the forked prefix too.
+/// TWO shapes carry it and both are live on disk, so neither branch may be
+/// dropped. The structured `source.subagent.thread_spawn.parent_thread_id` is
+/// the one every sub-agent rollout has (checked across on-disk rollouts from
+/// 0.117 through 0.147); the flat `parent_thread_id` is a later, additive
+/// mirror that only some versions write, always alongside the structured form
+/// and never on its own. Reading only the flat field — as this did before —
+/// therefore misses real sub-agents, which is what let their rollouts list as
+/// importable root sessions and let their replayed history be counted as their
+/// own (see [`is_forked_thread_header`]).
 ///
-/// So a forked thread yields no stats at all rather than wrong ones. The capsule
-/// still names the sub-agent and badges its thread id (both come from the
-/// PARENT's rollout); only the nested tool-call list is absent. The legacy
-/// `agent-<id>.jsonl` shape has no such prefix and is unaffected.
+/// A present-but-blank id is treated as absent: it names no parent, and a
+/// whitespace `parent_id` would still read as "this is a child" downstream.
 fn codex_parent_thread_id(payload: &serde_json::Value) -> Option<String> {
     payload
         .get("parent_thread_id")
@@ -1903,12 +1911,42 @@ fn codex_parent_thread_id(payload: &serde_json::Value) -> Option<String> {
         .map(ToOwned::to_owned)
 }
 
+/// Whether a transcript's opening record declares it a FORKED thread — codex
+/// 0.147's `fork_turns` sub-agent shape, where the child is a rollout of its own
+/// and the parent's history is copied into its head.
+///
+/// That copy is the problem: the parent's own tool calls sit at the top of the
+/// child's file, indistinguishable at the record level from the child's, so
+/// counting the whole file would credit the child with work the PARENT did. The
+/// hand-off that separates them (the addressed inter-agent `agent_message`) is
+/// not a reliable boundary either — a parent that already collected an earlier
+/// sibling's reply carries one inside the forked prefix too.
+///
+/// So a forked thread yields no stats at all rather than wrong ones. The capsule
+/// still names the sub-agent and badges its thread id (both come from the
+/// PARENT's rollout); only the nested tool-call list is absent. The legacy
+/// `agent-<id>.jsonl` shape has no such prefix and is unaffected.
+///
+/// BOTH markers are required, because "is a sub-agent" and "replays the parent"
+/// are different things and only the second one may be refused. A parent thread
+/// id alone says a thread was spawned by another; `forked_from_id` is codex's
+/// own declaration that this rollout was seeded with that thread's history. On
+/// disk they split cleanly — of 46 sub-agent rollouts, the 23 carrying
+/// `forked_from_id` are exactly the 23 that also replay a second `session_meta`
+/// header, and the other 23 are ordinary children whose tool calls are their
+/// own. Refusing on the parent id alone silently blanks the capsule for those.
 fn is_forked_thread_header(value: &serde_json::Value) -> bool {
-    value.get("type").and_then(|t| t.as_str()) == Some("session_meta")
-        && value
-            .get("payload")
-            .and_then(codex_parent_thread_id)
-            .is_some()
+    if value.get("type").and_then(|t| t.as_str()) != Some("session_meta") {
+        return false;
+    }
+    let Some(payload) = value.get("payload") else {
+        return false;
+    };
+    codex_parent_thread_id(payload).is_some()
+        && payload
+            .get("forked_from_id")
+            .and_then(serde_json::Value::as_str)
+            .is_some_and(|id| !id.trim().is_empty())
 }
 
 fn parse_codex_subagent_stats(
@@ -2168,6 +2206,7 @@ impl CodexParser {
         let mut messages = Vec::new();
         let mut cwd: Option<String> = None;
         let mut parent_id: Option<String> = None;
+        let mut session_header_seen = false;
         let mut git_branch: Option<String> = None;
         let mut model: Option<String> = None;
         let mut title: Option<String> = None;
@@ -2367,7 +2406,13 @@ impl CodexParser {
                             .get("cwd")
                             .and_then(|s| s.as_str())
                             .map(|s| s.to_string());
-                        parent_id = codex_parent_thread_id(payload);
+                        // First header wins — see the same latch in
+                        // `parse_jsonl_summary`: the replayed parent header that
+                        // follows a forked child's own declares no parent.
+                        if !session_header_seen {
+                            session_header_seen = true;
+                            parent_id = codex_parent_thread_id(payload);
+                        }
                         git_branch = payload
                             .get("git")
                             .and_then(|g| g.get("branch"))
@@ -4896,45 +4941,137 @@ mod tests {
         (temp_dir, parser, index_path)
     }
 
+    /// `fork_turns` copies the PARENT's history into the child's head, and that
+    /// copy carries the parent's own `session_meta` — on a real machine 23 of 46
+    /// sub-agent rollouts hold a second header. That header declares no parent
+    /// of its own, so reading `session_meta` last-record-wins clears the child's
+    /// `parent_id` and the rollout lists as an importable root again. The FIRST
+    /// header decides, exactly as `parse_codex_subagent_stats` already does via
+    /// its `checked_header` latch.
     #[test]
-    fn native_parent_thread_is_preserved_in_list_and_detail() {
+    fn replayed_parent_header_does_not_clear_the_child_parent_id() {
         let temp_dir = tempfile::tempdir().expect("create temp dir");
-        let conversation_id = "019ff3f7-937a-7671-b4f6-e404cb729d30";
-        let rollout_path = temp_dir
-            .path()
-            .join(format!("rollout-2026-08-28T10-00-00-{conversation_id}.jsonl"));
+        let child = "01a0098a-7e8a-72d3-b7c0-2df130c84063";
+        let parent = "01a0098a-5c58-7000-8000-000000000001";
         let lines = [
-            serde_json::json!({
-                "timestamp": "2026-08-28T10:00:00Z",
-                "type": "session_meta",
-                "payload": {
-                    "id": conversation_id,
+            rollout_line(
+                "2026-08-16T15:47:46Z",
+                "session_meta",
+                serde_json::json!({
+                    "id": child,
                     "cwd": "/tmp/demo",
-                    "parent_thread_id": " 019ff3f7-0000-7000-8000-000000000001 "
-                }
-            })
-            .to_string(),
-            serde_json::json!({
-                "timestamp": "2026-08-28T10:00:01Z",
-                "type": "event_msg",
-                "payload": {"type": "user_message", "message": "child task"}
-            })
-            .to_string(),
+                    "source": {"subagent": {"thread_spawn": {"parent_thread_id": parent}}}
+                }),
+            ),
+            // The replayed parent header, verbatim as codex writes it.
+            rollout_line(
+                "2026-08-16T15:47:46Z",
+                "session_meta",
+                serde_json::json!({"id": parent, "cwd": "/tmp/demo"}),
+            ),
+            rollout_line(
+                "2026-08-16T15:47:47Z",
+                "event_msg",
+                serde_json::json!({"type": "user_message", "message": "child task"}),
+            ),
         ];
-        fs::write(&rollout_path, format!("{}\n", lines.join("\n"))).expect("write rollout");
+        fs::write(
+            temp_dir
+                .path()
+                .join(format!("rollout-2026-08-16T15-47-46-{child}.jsonl")),
+            format!("{}\n", lines.join("\n")),
+        )
+        .expect("write rollout");
 
         let parser = CodexParser::with_base_dir(temp_dir.path().to_path_buf());
         let summaries = parser.list_conversations().expect("list conversations");
         assert_eq!(summaries.len(), 1);
         assert_eq!(
             summaries[0].parent_id.as_deref(),
-            Some("019ff3f7-0000-7000-8000-000000000001")
+            Some(parent),
+            "a replayed parent header must not clear the child's parent_id"
         );
 
         let detail = parser
-            .get_conversation(conversation_id)
+            .get_conversation(child)
             .expect("load conversation detail");
-        assert_eq!(detail.summary.parent_id, summaries[0].parent_id);
+        assert_eq!(detail.summary.parent_id.as_deref(), Some(parent));
+    }
+
+    /// Both on-disk sub-agent shapes must surface `parent_id`, because that is
+    /// the only thing keeping these rollouts out of the importer's root list
+    /// (`import_service::collect_local_summaries` drops `parent_id.is_some()`).
+    /// The STRUCTURED case is the one that matters most: it is the shape every
+    /// real sub-agent rollout carries, while the flat mirror only accompanies
+    /// it on some versions.
+    #[test]
+    fn native_parent_thread_is_preserved_in_list_and_detail() {
+        let parent = "019ff3f7-0000-7000-8000-000000000001";
+        let temp_dir = tempfile::tempdir().expect("create temp dir");
+
+        // Same parent id, two payload shapes, one directory — so a regression in
+        // either branch shows up as a listed root session.
+        let cases = [
+            (
+                "019ff3f7-937a-7671-b4f6-e404cb729d30",
+                serde_json::json!({ "parent_thread_id": format!(" {parent} ") }),
+            ),
+            (
+                "019ff3f7-937a-7671-b4f6-e404cb729d31",
+                serde_json::json!({
+                    "source": {"subagent": {"thread_spawn": {
+                        "parent_thread_id": parent,
+                        "depth": 1,
+                        "agent_nickname": "Gibbs",
+                        "agent_role": "worker"
+                    }}}
+                }),
+            ),
+        ];
+
+        for (conversation_id, extra) in &cases {
+            let mut payload = serde_json::json!({"id": conversation_id, "cwd": "/tmp/demo"});
+            let map = payload.as_object_mut().expect("payload object");
+            for (k, v) in extra.as_object().expect("extra object") {
+                map.insert(k.clone(), v.clone());
+            }
+            let lines = [
+                rollout_line("2026-08-28T10:00:00Z", "session_meta", payload),
+                rollout_line(
+                    "2026-08-28T10:00:01Z",
+                    "event_msg",
+                    serde_json::json!({"type": "user_message", "message": "child task"}),
+                ),
+            ];
+            fs::write(
+                temp_dir
+                    .path()
+                    .join(format!("rollout-2026-08-28T10-00-00-{conversation_id}.jsonl")),
+                format!("{}\n", lines.join("\n")),
+            )
+            .expect("write rollout");
+        }
+
+        let parser = CodexParser::with_base_dir(temp_dir.path().to_path_buf());
+        let summaries = parser.list_conversations().expect("list conversations");
+        assert_eq!(summaries.len(), cases.len());
+
+        for (conversation_id, _) in &cases {
+            let listed = summaries
+                .iter()
+                .find(|s| s.id == *conversation_id)
+                .unwrap_or_else(|| panic!("{conversation_id} listed"));
+            assert_eq!(
+                listed.parent_id.as_deref(),
+                Some(parent),
+                "{conversation_id} must report its parent so the importer skips it"
+            );
+
+            let detail = parser
+                .get_conversation(conversation_id)
+                .expect("load conversation detail");
+            assert_eq!(detail.summary.parent_id, listed.parent_id);
+        }
     }
 
     #[test]
@@ -7739,9 +7876,13 @@ mod tests {
     #[test]
     fn forked_child_transcript_yields_no_stats() {
         // A codex 0.147 sub-agent thread opens with its own `session_meta`
-        // (carrying `parent_thread_id`) and then replays the PARENT's history,
-        // tool calls included. Counting those would credit the child with the
-        // parent's work, so the whole file is refused.
+        // (carrying `parent_thread_id` AND `forked_from_id`) and then replays the
+        // PARENT's history, tool calls included. Counting those would credit the
+        // child with the parent's work, so the whole file is refused.
+        //
+        // `forked_from_id` is what makes it a replay, and on disk it is never
+        // absent from one: the 23 sub-agent rollouts that carry it are exactly
+        // the 23 that also hold the replayed second header below.
         let child = "01a0098a-7e8a-72d3-b7c0-2df130c84063";
         let dir = temp_session_dir("forked-child");
         // The lookup matches a rollout whose stem ends with the thread id,
@@ -7752,7 +7893,10 @@ mod tests {
                 rollout_line(
                     "2026-08-16T07:47:46Z",
                     "session_meta",
-                    serde_json::json!({"id":child,"parent_thread_id":"parent","cwd":"/tmp/demo"}),
+                    serde_json::json!({
+                        "id": child, "cwd": "/tmp/demo",
+                        "parent_thread_id": "parent", "forked_from_id": "parent"
+                    }),
                 ),
                 rollout_line(
                     "2026-08-16T07:47:46Z",
@@ -7776,6 +7920,84 @@ mod tests {
             "a forked child transcript must contribute no stats"
         );
 
+        // The same replay as codex actually writes it: the parent id lives under
+        // the structured subagent source and there is NO flat mirror. Matching
+        // only the flat field counted these files, crediting the child with the
+        // parent's replayed tool calls.
+        let structured = "019e88cb-ada0-7611-b7cf-25a6e3535722";
+        fs::write(
+            dir.join(format!("rollout-2026-06-02T22-45-10-{structured}.jsonl")),
+            [
+                rollout_line(
+                    "2026-06-02T22:45:10Z",
+                    "session_meta",
+                    serde_json::json!({
+                        "id": structured,
+                        "cwd": "/tmp/demo",
+                        "forked_from_id": "parent",
+                        "source": {"subagent": {"thread_spawn": {
+                            "parent_thread_id": "parent", "depth": 1, "agent_role": "explorer"
+                        }}}
+                    }),
+                ),
+                rollout_line(
+                    "2026-06-02T22:45:10Z",
+                    "session_meta",
+                    serde_json::json!({"id": "parent", "cwd": "/tmp/demo"}),
+                ),
+                rollout_line(
+                    "2026-06-02T22:45:11Z",
+                    "response_item",
+                    serde_json::json!({
+                        "type":"function_call","call_id":"parents_own","name":"exec_command",
+                        "arguments": serde_json::json!({"cmd":"git status"}).to_string(),
+                    }),
+                ),
+            ]
+            .join("\n"),
+        )
+        .expect("write structured forked child");
+        assert!(
+            parse_codex_subagent_stats(&dir, structured).is_none(),
+            "the structured subagent shape must be refused too"
+        );
+
+        // A spawned child that was NOT seeded with the parent's history: it has a
+        // parent thread id but no `forked_from_id` and no replayed header, so the
+        // tool calls in it are its own and MUST still be counted. Half the
+        // sub-agent rollouts on disk look like this; refusing them on the parent
+        // id alone blanks the capsule's tool list for real work.
+        let clean = "019d9929-6a00-7543-b045-172ebb06e5eb";
+        fs::write(
+            dir.join(format!("rollout-2026-04-17T01-58-42-{clean}.jsonl")),
+            [
+                rollout_line(
+                    "2026-04-17T01:58:42Z",
+                    "session_meta",
+                    serde_json::json!({
+                        "id": clean,
+                        "cwd": "/tmp/demo",
+                        "source": {"subagent": {"thread_spawn": {
+                            "parent_thread_id": "parent", "depth": 1, "agent_role": "worker"
+                        }}}
+                    }),
+                ),
+                rollout_line(
+                    "2026-04-17T01:58:43Z",
+                    "response_item",
+                    serde_json::json!({
+                        "type":"function_call","call_id":"its_own","name":"exec_command",
+                        "arguments": serde_json::json!({"cmd":"pnpm test"}).to_string(),
+                    }),
+                ),
+            ]
+            .join("\n"),
+        )
+        .expect("write clean child");
+        let clean_stats =
+            parse_codex_subagent_stats(&dir, clean).expect("a non-replayed child keeps its stats");
+        assert_eq!(clean_stats.tool_calls.len(), 1);
+
         // The legacy shape has no replayed prefix and still resolves.
         fs::write(
             dir.join("agent-solo.jsonl"),
```

</details>

TroyMitchell911:
Resolved in fb7d04e7: parent IDs now latch on the first header, replay suppression requires forked_from_id, and flat/structured replay plus clean-child regressions are covered. Full local CI passes.

PR #607, opened 2026-08-30
fix(cursor-acp): enable parameterized model picker

ZY-Dragon03:
## Problem

Cursor ACP only exposes parameterized configuration such as the Fast selector when the client advertises `_meta.parameterizedModelPicker=true`.

Without this capability, Codeg falls back to the default Cursor model variant behavior and Composer 2.5 may remain on Fast.

## Fix

Advertise `parameterizedModelPicker` for agents whose resolved ACP backend launches `cursor-agent ... acp`, covering both the built-in Cursor agent and custom Cursor-backed agents.

Detection uses the resolved launch command and arguments, not custom agent names.

## Verification

- built-in Cursor receives the capability
- custom `cursor-agent ... acp` agents receive the capability
- non-Cursor custom ACP agents do not
- Cursor exposes `mode`, `model`, and `fast`
- `session/set_config_option` successfully applies `fast=false`

This matches the compatibility approach used by other ACP clients that support Cursor's parameterized model picker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,9 @@ src-tauri/binaries/
 
 # pnpm local store (pnpm 11+)
 .pnpm-store/
+
+# `rustc foo.rs` with no -o writes its binary here. Scratch compiles are how
+# you check a borrow or a lifetime question in isolation, and the leftover
+# executable then sits untracked in the repo root — which is enough to fail a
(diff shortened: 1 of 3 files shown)


xintaofei:
codeg work task `51` is done — https://github.com/xintaofei/codeg/pull/607 (3 files, +198/-21).

PR #634, opened 2026-09-02
feat(acp): bump claude-agent-acp to 0.73.0

Adam-Dalloul:
Bumps the pinned Claude ACP adapter from 0.69.0 to 0.73.0.

Why: Anthropic ships new model availability through the bundled runtime. The raw model id `claude-fable-5-1` is live, but the API rejects it on older runtimes with "Claude Code 2.1.247 does not support this model; version 2.1.251 or newer is required". The 0.69.0 pin bundles claude-agent-sdk 0.3.232 (runtime 2.1.232); 0.73.0 bundles 0.3.257 (runtime 2.1.257), so new Anthropic models surface in the model config option with no code change.

Verified live over ACP stdio: 0.73.0 boots, answers initialize, and session/new succeeds and advertises the model config option (Default / Opus 1M / Fable / Sonnet / Haiku).

Tarball-diffed all four releases in between. The per-version analysis (0.70.0 through 0.73.0) is in the registry comment above the pin, where it lives for every other version; comments that described the old pin's behavior (the option-level permission `changes[]` claim, the AIR tripwire enumeration, the title fallback) are updated in the same change.

### DIFF
--- a/src-tauri/src/acp/background_watch.rs
+++ b/src-tauri/src/acp/background_watch.rs
@@ -39,13 +39,19 @@
 //!
 //! * **Session title** — Claude Code's generated name arrives as a dedicated
 //!   `ai-title` transcript record whenever the background summarizer finishes,
-//!   routinely AFTER the turn that triggered it ended. The ACP adapter only
-//!   pulls the name at turn-end (`maybeUpdateSessionTitle`, claude-agent-acp
-//!   0.69.0), so on a short session there is nothing to read yet and no wire
-//!   event ever follows. These bytes are already being tailed, so the records
-//!   are folded here and handed to [`publish_native_title`] — the same path a
-//!   live ACP title takes. Not activity: it rides alongside the activity event
-//!   rather than inside it (see `run_watch`).
+//!   routinely AFTER the turn that triggered it ended. Through claude-agent-acp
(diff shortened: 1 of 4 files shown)


Adam-Dalloul:
Superseded by main — closing is the right call.

`6bc1c1f6` already moved the pin to `0.73.0` (registry.rs and the preflight test both read `@agentclientprotocol/claude-agent-acp@0.73.0`), and the 0.70.0–0.73.0 analysis on main is both fuller and more current than what this branch carries. Main also went further than this PR did: it actually adopts `asyncTasks` in `build_client_capabilities`, where this branch only documented it as not-adopted — so rebasing would re-assert a paragraph main has already disproved.

One small leftover, unrelated to the pin: the `background_watch.rs` module doc still says the adapter "only pulls the name at turn-end (`maybeUpdateSessionTitle`, claude-agent-acp 0.69.0)", which the `session-titles.js` note in registry.rs now contradicts. Happy to send that one-paragraph doc fix separately if it's wanted.

Adam-Dalloul:
Closing: 6bc1c1f6 already moved the pin to 0.73.0 with your own per-version analysis, and main goes further by actually adopting asyncTasks. Nothing left here worth rebasing.

PR #635, opened 2026-09-02
feat(acp): let an agent opt into the latest adapter release

Adam-Dalloul:
New Anthropic models ship inside the adapter's bundled runtime, so they can sit behind the pin. claude-fable-5-1 is live today, but through the pinned claude-agent-acp 0.69.0 (bundled runtime 2.1.232) every request to it answers:

```
API Error: 400 Claude Code 2.1.232 does not support this model; version 2.1.251 or newer is required. Run 'claude update', or update the Claude desktop app, then try again.
```

The 0.73.0 adapter bundles 2.1.257 and the model just works. Until a pin moves, a user who wants such a model has no supported way to get it.

This adds an opt-in, per-agent "Adapter version" control in Agent Settings: Pinned (recommended) or Latest (unreviewed). It is generic across every npx-distributed agent and hidden for binary/uvx agents, and it holds three properties:

- The default is unchanged. Pinned stays the default, labeled recommended, and the pinned install spec is byte-identical to today's. The copy under the control warns that the newest releases are unreviewed and can break the agent, and that one is occasionally known broken, the way Kimi Code 0.37.x was.
- Fail safe to the pin. Latest is resolved only at install and upgrade time, with the same npm flags installs already use (--include=optional, --registry). If installing @latest fails because npm is unreachable or a mirror lags, the install log says so and the pinned spec is installed instead, so the agent still works. Launch never consults npm.
- Honest version reporting. The installed version keeps coming from the real post-install probe, so Version Status shows what actually landed; on the latest channel the pass state keeps Upgrade available, since codeg cannot know whether npm has something newer.

Storage follows the pi runtime-override pattern: a reserved env_json key (CODEG_ADAPTER_CHANNEL=latest) behind a structured control, saved through the existing env save path, so it works in desktop and server modes with no schema change. Background update checks, prompts and telemetry were deliberately left out; nothing polls npm. An explicit Custom install version still wins on either channel and never falls back.

Orthogonal to #634: this changes no pins.

Rust helpers are unit-tested (pinned default, latest mapping, override wins, the env sentinel), the frontend helpers and version-check states too, and the strings are translated in all ten locales. tsc, eslint, prettier and the full vitest suite pass locally.


xintaofei:
codeg work task `142` is done — https://github.com/xintaofei/codeg/pull/635 (1 file, +20/-0).

PR #636, opened 2026-09-02
fix(chat): show a message sent mid-turn as a message, not a strip

Adam-Dalloul:
Messages sent while a turn is streaming are delivered to the agent, but the live view never shows them as messages.

Two things go wrong. The strip above the composer flips to "received" and then sits there for the rest of the turn, so a sent message never appears as one. And because no user turn lands between them, the reply to that message continues inside the same assistant bubble as the previous one, gluing two separate answers together mid-paragraph ("...I'll report both links once CI is green.Not done - those are the two PRs...").

A note that arrives already delivered came in over the native `_session/steering` push, which means the agent has the text as a user message. It now becomes a `steering` block on the live message: the assistant turn closes there, the message renders as its own user turn, and the reply to it starts a new turn. This is what `project_turns` in `parsers/acp_native.rs` already does with a mid-turn `user_message_chunk`, so the live view now matches what you get on reload. A detail fetch that lands mid-turn and brings back the agent's own copy is matched on content, keeping the live copy, since only that one sits between the two halves of the reply.

Pending notes are unchanged. The `check_user_feedback` pull channel reaches the agent as a tool result and never as a user message, so it has no user turn on reload either and stays a strip. Queued messages, which are only sent once the turn ends, are untouched. A note that arrives with no turn running is not adopted and keeps its strip, so nothing can end up showing in neither place.


xintaofei:
codeg work task `143` is done — https://github.com/xintaofei/codeg/pull/636 (99 files, +8942/-698).

PR #638, opened 2026-09-02
fix(acp): preserve UNC cwd for Windows batch launchers

damiandelmas:
## Summary

- detect Windows UNC workspaces when the resolved ACP agent is a `.cmd` or `.bat` launcher
- launch that narrow case through the system `cmd.exe` and `pushd`, avoiding cmd's fallback to `C:\Windows`
- preserve the existing direct-spawn behavior for local paths, native executables, and non-Windows hosts
- defensively encode batch arguments using Rust stdlib-compatible rules; disable AutoRun and delayed expansion

## Why

A native Windows Codeg session can open a WSL workspace such as `\\wsl.localhost\Ubuntu\home\user\repo`. Codeg correctly passes that project path as the ACP process cwd. For npm-backed agents such as Hermes, however, the resolved launcher is a Windows `.cmd` file. `cmd.exe` rejects a UNC process cwd and silently falls back to the Windows directory, so the ACP session reports the project while the agent process derives its terminal cwd from `C:\Windows`.

`pushd` maps the UNC share to a temporary drive for the lifetime of the command processor, retaining the intended workspace without changing agent installation or execution namespaces.

## Validation

- `cargo test --manifest-path vendor/sacp-tokio/Cargo.toml --lib` — 16 passed
- `cargo check --manifest-path vendor/sacp-tokio/Cargo.toml --lib --target x86_64-pc-windows-gnu` — passed
- `git diff --check` — passed

A live UNC integration run still requires a Windows host with an accessible UNC fixture; the existing CI Windows matrix will compile the Windows-specific path.

### DIFF
--- a/src-tauri/vendor/sacp-tokio/src/acp_agent.rs
+++ b/src-tauri/vendor/sacp-tokio/src/acp_agent.rs
@@ -15,6 +15,106 @@ use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 const CREATE_NO_WINDOW: u32 = 0x0800_0000;
 const MAX_STDERR_CAPTURE_BYTES: usize = 1024 * 1024;
 
+fn is_windows_unc_path(path: &str) -> bool {
+    let lower = path.to_ascii_lowercase();
+    lower.starts_with(r"\\?\unc\")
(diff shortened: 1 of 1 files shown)


xintaofei:
codeg work task `145` is done — https://github.com/xintaofei/codeg/pull/638 (1 file, +201/-13).

PR #640, opened 2026-09-02
fix(chat): steer a draft with image attachments instead of refusing it

Adam-Dalloul:
With the Claude Code agent, "Insert into current turn" refused any draft that had an image staged: the menu entry disabled itself with a text-only tooltip, and a draft holding file badges was quietly rerouted to the queue. The limit was on our side, not the wire's. The steering chain narrowed everything to a bare string (`ConnectionCommand::Steer { text }`, `build_steer_params` hardcoding a single text block), while claude-agent-acp's `_session/steering` handler feeds the prompt array through the exact same conversion as `session/prompt`, image blocks included.

Steer now carries the draft's `PromptInputBlock`s end to end. The composer sends the full block list whenever the draft holds more than plain text, with the display text as the recorded note; the backend maps it with `map_prompt_blocks` (the `session/prompt` encoding) and re-hydrates uploaded `file://` markers exactly like `/acp_prompt`, so web and remote mode work unchanged. The prompt ledger fingerprints the steered blocks the way a prompt's are, and a text-only steer stays byte-identical to before.

Attachments are still never silently dropped: the pull tool carries plain text only, so a blocks-bearing note that races a mid-session channel downgrade is rejected with `NoActiveTurn` and the composer's existing fallback queues the whole draft, image included, with the usual toast. Steering while an upload is still settling gets the same "still uploading" toast a plain send does, and the now-unused `steerAttachmentsUnsupported` string is removed from all ten locales.


xintaofei:
codeg work task `146` is done — https://github.com/xintaofei/codeg/pull/640 (97 files, +8431/-659).

PR #643, opened 2026-09-02
fix(acp): match a slash command against the prompt ledger it was rebuilt from

Adam-Dalloul:
Sending `/goal <objective>` renders the whole turn twice while it streams: the prompt bubble and its reply both appear a second time, one copy completed and one still running.

The background watcher tells a turn codeg sent apart from out-of-turn work by matching a transcript record's initiator text against the prompt ledger. A slash command has no readable initiator text, so `slash_command_display` rebuilds it from the command tags as the name, one space, and the trimmed args. That drops whatever separator the sender actually typed, and the composer already inserts a space after a command badge, so a sender who types their own sends two. The rebuilt text no longer starts with the fingerprint, the command classifies as an out-of-turn initiator, and the wire-rendered turn comes back as a background overlay beside itself. Same duplicate that the submission window already covers for a command's side records, reached through the command record itself.

The match now also runs over collapsed whitespace, which stays exact in words and order. Two tests: the end-to-end goal turn sent with an extra space must not surface an overlay, and the ledger must still refuse different words and still consume an entry exactly once.

### DIFF
--- a/src-tauri/src/acp/background_watch.rs
+++ b/src-tauri/src/acp/background_watch.rs
@@ -112,6 +112,20 @@ const MAX_EPISODE_MESSAGES: usize = 512;
 /// boundary rotation always wins for multi-turn episodes.
 const FORCE_ROTATE_MESSAGES: usize = MAX_EPISODE_MESSAGES * 2;
 
+/// Every run of whitespace as a single space, ends trimmed. Used by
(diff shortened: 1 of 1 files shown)


xintaofei:
codeg work task `150` is done — https://github.com/xintaofei/codeg/pull/643 (1 file, +111/-49).

PR #646, opened 2026-09-03
fix(mcp): isolate local scan source failures

### LINKED ISSUE #632: 【bug】本地 MCP 只解析 .gemini\config\mcp_config.json，空文件即整体加载失败、已配置的 MCP 全不显示
## 环境
   - codeg-server: v0.29.0
   - 系统: Windows x64

## 问题内容
   偏好设置 → MCP → 本地 MCP，直接报红加载失败，并显示"当前未检测到本地 MCP"，点刷新无效：

 ```
 加载失败：invalid JSON at C:\Users<用户名>.gemini\config\mcp_config.json:
 EOF while parsing a value at line 1 column 0
```
Image

但是我在pi里装了好几个mcp了

Image

也让AI看了下
```
   1. 报错指向的 `C:\Users\<用户名>\.gemini\config\mcp_config.json` 是一个 **0 字节空文件**，所以 JSON 解析一进来就 EOF
 报错。
   2. codeg 把"本地 MCP"的唯一来源指到了 **Gemini CLI 的配置路径**，一旦这个文件为空/损坏，就让整个本地 MCP 列表加载失
 败、一个都不显示。
```

   我本机其实配了很多 MCP，codeg 一个都没读到：
   - `C:\Users\<用户名>\.claude.json` 里有 8 个：`github / filesystem / sequential-thinking / exa / memory / context7 /
 tinyfish / blender`

   ## 复现
   1. 确保 `~/.gemini/config/mcp_config.json` 存在且为空文件（0 字节）。
   2. 打开 codeg → MCP → 本地 MCP。
   3. 直接报 "invalid JSON ... EOF"，列表全空，刷新无效。

   ## 期望
   1. 空文件 / 损坏文件应当作"无配置"跳过，而不是让整个本地 MCP 加载失败。
   2. 单个来源解析失败要**降级**：跳过坏的、继续显示能读到的，并把错误做成某一项的告警而非整页失败。
   3. 本地 MCP 最好能从多个常见来源聚合（至少别把用户在 `.claude.json` 等处已配置的 MCP 全漏掉），或提供可配置的扫描路
 径。

dawNotPoi:
## Summary

- continue local MCP discovery when one application config cannot be read or parsed
- keep server lookup and all mutation paths strict so writes never use an incomplete view
- preserve source priority, app union, and the OpenClaw/Kimi merge rules

## Testing

- cargo test --no-default-features --lib commands::mcp::tests::
- cargo test --no-default-features --lib
- cargo check --no-default-features --lib --bin codeg-server
- cargo clippy --no-default-features --lib --bin codeg-server -- -D warnings
- cargo check --no-default-features --bin codeg-mcp
- cargo clippy --no-default-features --bin codeg-mcp -- -D warnings

Closes #632


xintaofei:
Thanks for this — the reader table is a real improvement over the 15 copy-pasted loops, and I kept it as-is. I'd independently worked on #632 before seeing this PR, so rather than open a competing one I've pushed my work on top of your commit (`bf71cdde`, fast-forward, your commit untouched underneath). Happy to discuss any of it.

**What I kept from your commit:** `local_mcp_readers()` and the single generic loop. Adding a 15th agent is now one line instead of another copy-paste block. Your `successful_scan_preserves_openclaw_and_kimi_merge_rules` test stays too — it covers a merge rule that had no test before.

**What changed, and why.**

1. **An empty config is not a corrupt one.** #632's file is a 0-byte `~/.gemini/config/mcp_config.json` — Antigravity touches it into existence before it ever writes a server, so `EOF while parsing a value at line 1 column 0` is exactly what a *healthy* uninitialised install produces. Isolating the source fixes the list, but that agent then stays invisible forever *and* unwritable: every writer starts with `read_json_file` on the file it is about to change (e.g. `upsert_antigravity_server_at`), so nothing can put content into it again. The file is empty → the write is refused → the file stays empty. `read_json_file` now treats empty and whitespace-only as `{}`, which is the rule the Hermes reader already applied to its own config.

2. **Two sources could fail without the scan noticing**, which made the fail-closed half of the isolation a no-op:
   - `read_hermes_servers` swallowed I/O errors *and* invalid YAML into `Ok(empty)`, so Hermes went silently absent rather than reporting a failure. Strict mode then believed the scan was complete.
   - Every JSON reader plus the Codex and Grok TOML readers gated on `if !path.exists()`. `Path::exists()` is `false` for *any* failed stat — a permission wall or a symlink loop reads as "not configured". New `read_config_to_string` makes the read itself the test and carries the path into the error detail. (Verified with a symlinked `~/.gemini/settings.json` loop: now surfaces `Too many levels of symbolic links (os error 62)` instead of silently reporting the agent as unconfigured.)

3. **`Strict` mode replaced by an explicit write guard.** The scan is now infallible and returns `LocalMcpScan { servers, warnings }`; `mcp_upsert_local_server` and `mcp_set_server_apps` call `require_complete_scan` *before* their first write. Under `Strict`, `mcp_upsert_local_server`'s `find_local_server` ran **after** the write loop as a reload — so a degraded scan reported failure on a save that had already landed in fourteen configs.

4. **The warnings are surfaced, not just logged.** `tracing::warn!` is invisible to the person whose config is broken; they just see an agent quietly missing from the list. The settings page now names the agent and the reason, and disables Save/Create while any source is degraded — otherwise a draft seeded from a partial scan could be submitted after an out-of-band repair, carrying a stale app list that removes the server from agents the user never unchecked.

**Verification** (all green on `bf71cdde`):
- `cargo clippy --all-targets --features test-utils -- -D warnings`
- `cargo test --features test-utils` — 3354 passed, 0 failed
- `cargo test --no-default-features --bin codeg-server --lib` — 3323 passed, 0 failed
- `cargo clippy --no-default-features --bin codeg-mcp -- -D warnings`
- `tsc --noEmit`, `eslint`, `prettier --check`, `vitest run` (402 files / 5746 tests), `pnpm build`

Plus an end-to-end check against a real `codeg-server` with a fake `$HOME`: 0-byte config → 200 with all servers and no warning; corrupt config → 200 with servers plus a warning naming the file, 422 on save with the other configs provably unmutated, save succeeds once repaired.

One assertion of yours is inverted: `empty_antigravity_file_remains_a_reader_error` → `empty_antigravity_file_reads_as_no_servers`, for the reason in (1). `strict_scan_still_fails_on_a_single_broken_source` became `a_failed_source_is_refused_by_the_write_guard`, testing the same invariant through the guard that replaced the mode flag.

PR #647, opened 2026-09-03
feat(terminal): mobile virtual key bar

mrytsr:
## What

Add an on-screen virtual key bar for the terminal panel so mobile users can send escape / tab / control / arrow / page keys without a hardware keyboard.

## Why

Without a physical keyboard, mobile terminal usage is essentially broken: no ESC to leave insert mode in vim, no Ctrl+C for SIGINT, no arrow keys for readline history. The mobile Drawer container plus a two-row button grid closes that gap.

## Design

- **Pure encoding** (`lib/terminal/keybar.ts`): 12 keys + `termKeySeq` + `applyTermMods` + `kbdLiftPx`. 18 unit tests.
- **`<TermKeybar />`**: presentational, shadcn Button + Tailwind, no raw CSS. 5 component tests.
- **Write-path integration**: the key bar pushes bytes into the existing ordered single-flight write pump (`writeQueueRef`), not `terminalWrite` directly — otherwise concurrent soft-keyboard and key-bar presses would scramble under Tauri invoke reordering.
- **Latch semantics**: CTRL/ALT are one-shot sticky keys. The next `onData` (or the next key-bar press) wraps itself then auto-clears. Refs hold the truth, state drives the highlight.
- **Soft-keyboard follow**: `visualViewport` listener computes `kbdLift = max(0, innerHeight - vvHeight - vvOffsetTop)` and applies `max-height: calc(100% - kbdLift)` to the outer flex column so the bar lifts above the keyboard and xterm auto-refits via the existing ResizeObserver.
- **Collapse state**: persisted to `localStorage codeg:term-keybar` at the panel level (not per tab) so switching tabs stays consistent.
- **Desktop untouched**: `useMediaQuery("(max-width: 768px)")` short-circuits both the toggle button and the bar — desktop users see no change.

## Verification

- `pnpm test`: 5764 passed (includes 23 new keybar tests)
- `pnpm eslint .`: clean
- `pnpm build`: 32/32 static pages generated
- `tsc --noEmit`: clean

## i18n

16 new keys under `Folder.terminal.keybar` across all 10 locales. The symbols (ESC/TAB/↑↓←→ etc.) are language-neutral; `show` / `hide` / `label` are translated for zh-CN/zh-TW/ja/ko/es/de/fr/pt/ar.

## Side fix

Mobile Drawer height bumped from 70vh to 95vh to give the new bar enough room when folded in.

### DIFF
--- a/src/app/workspace/layout.tsx
+++ b/src/app/workspace/layout.tsx
@@ -683,7 +683,7 @@ function MobileFolderWorkspaceShell({
         swipeDirection="down"
         disablePointerDismissal={false}
       >
-        <DrawerContent showCloseButton={false} className="h-[70vh] p-0">
+        <DrawerContent showCloseButton={false} className="h-[95vh] p-0">
           <DrawerTitle className="sr-only">Terminal</DrawerTitle>
           <div className="h-full min-h-0 overflow-hidden">
             <TerminalPanel />
--- a/src/components/terminal/term-keybar.test.tsx
+++ b/src/components/terminal/term-keybar.test.tsx
@@ -1,0 +1,99 @@
+import { fireEvent, render, screen } from "@testing-library/react"
+import { beforeEach, describe, expect, it, vi } from "vitest"
+
+import { TermKeybar } from "./term-keybar"
+
+// next-intl is mocked at module level via vitest config; provide the minimum
+// shape used by `useTranslations` so `t("...")` returns the key path.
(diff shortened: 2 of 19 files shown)


xintaofei:
codeg work task `154` is done — https://github.com/xintaofei/codeg/pull/647 (17 files, +330/-79).

PR #652, opened 2026-09-04
fix(codex): treat empty grep exit 1 as no matches

### LINKED ISSUE #651: Codex rg searches with no matches are rendered as failed tool calls
## Summary

When a Codex agent runs `rg` and the search finds nothing, Codeg shows the terminal tool call as failed, in red. Nothing failed. Ripgrep returns exit code 1 to mean "no matches were found."

I have hit this in conversation after conversation. A successful negative search looks like a broken command, and the agent reads it the same way I do, so it retries the search or falls back to slower file scanning.

The root cause is upstream, not in Codeg. Codeg is still the layer where a fix reaches users soonest.

## Environment

- Codeg Desktop 0.30.2
- Windows
- PowerShell
- Codex agents through ACP

## Reproduction

1. Open a Codex conversation in Codeg.
2. Search for a token that does not exist:

```powershell
rg -n "__definitely_absent_token__" README.md
```

3. The command writes nothing to stdout or stderr. Codeg labels the tool call failed.

A captured ACP transcript contains this terminal result:

```json
{
  "_meta": {
    "terminal_exit": {
      "exit_code": 1,
      "signal": null
    }
  },
  "rawOutput": {
    "exit_code": 1,
    "formatted_output": ""
  },
  "status": "failed"
}
```

## Repeated examples

I reproduced this in four separate Codeg conversations:

- Conversation 117: `rg -n -C 50 '"id": "CP-0090"|"id": "CP-0085"|"id": "CP-0084"' docs/backlog.json` returned no output, marked failed.
- Conversation 113: `rg -n -i "Reconcile post-B-2 accessibility delta|post-B-2 accessibility|..." docs .github 2>$null` returned no output, marked failed.
- Conversation 90: a bounded `rg` search across the docs returned no output, marked failed.
- Conversation 103: the compound case. An initial `rg --files` printed a valid filename, a later `rg` found nothing, and Codeg rendered the whole tool call as failed because the last process exited 1.

Real ripgrep errors look different. The same transcripts contain malformed regexes, missing paths, and invalid Windows globs, and every one of those wrote a diagnostic to stderr. Those should stay failures.

## Why it happens

Ripgrep's documented exit contract is:

- `0`: a match was found
- `1`: no matches were found
- `2`: an actual error occurred

Codex reports a shell command as failed whenever its exit code is nonzero. openai/codex#8268 reports the same behavior upstream, and it is closed as not planned.

Codeg keeps that wire status as-is. In `src-tauri/src/parsers/acp_native.rs` the parser copies the incoming status and treats `status == "failed"` as `is_error`. The UI has no neutral state left to show for an expected empty result.

## Expected behavior

Show an `rg` run that exits 1 with no matches neutrally. Something like:

- `No matches found`, or
- `Completed, no matches`

It should not get the same red failure treatment as a malformed regex, a missing path, a permission denial, or a process that never launched.

## Possible approaches

ACP hands Codeg a `failed` status, so any fix here normalizes at the integration or presentation layer. Only when all of these hold:

- the executed command is a standalone ripgrep search;
- the exit code is 1;
- there is no stderr or error diagnostic;
- the process was not canceled or denied.

Another option: a neutral nonzero-exit state such as `Exited 1, no output`. Or let a command declare which exit codes count as success, without rewriting the underlying ACP record.

Please do not treat every exit code 1 as success. Real errors and compound shell commands need to keep their actual meaning. Regression tests should separate four cases:

1. `rg` with matches, which is a success;
2. `rg` with no matches and no diagnostics, which is a neutral no-match;
3. malformed regex or missing path, which stays a failure;
4. compound and pipelined commands, which keep the overall shell result or get shown without claiming every command in the chain failed.

## Impact

- Misleading red failure markers in otherwise healthy agent runs.
- Agents read a valid negative search as a broken command or a broken environment.
- Wasted retries and slower fallback scans.
- When every fourth tool call is red, you stop reading the red.

dawNotPoi:
## Summary

- treat the exact Codex grep no-match result (`exit_code: 1` with empty output) as a completed presentation state
- preserve the original ACP result and exit code while keeping real command, glob, diagnostic, and higher-exit failures on the error path
- cover live/reloaded and ID/position-paired tool results with positive and negative regression tests

Closes #651

## Validation

- `pnpm test` (402 files, 5760 tests)
- `pnpm eslint .`
- `pnpm exec tsc --noEmit`
- `pnpm build`
- `pnpm exec prettier --check src/lib/adapters/ai-elements-adapter.ts src/lib/adapters/ai-elements-adapter.test.ts`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/src/components/message/content-parts-renderer.tsx
+++ b/src/components/message/content-parts-renderer.tsx
@@ -56,7 +56,10 @@ import {
 } from "./context-compaction-card"
 import { FeedbackCheckResultCard } from "./feedback-check-result-card"
 import { SearchResultsOutput } from "./search-results-output"
-import { parseCodexCommandEnvelope } from "@/lib/codex-command-action"
+import {
+  isCodexGrepNoMatchEnvelope,
+  parseCodexCommandEnvelope,
+} from "@/lib/codex-command-action"
 import {
   CODEX_SCRIPT_TOOL_NAME,
   parseCodexScriptCard,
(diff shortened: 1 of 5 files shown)


xintaofei:
codeg work task `156` is done — https://github.com/xintaofei/codeg/pull/652 (5 files, +100/-21).

PR #655, opened 2026-09-04
feat(acp): move codex-acp to 1.10.0

Adam-Dalloul:
OpenAI shipped `gpt-6-astra`, and the pinned adapter cannot offer it.

The model list comes from the codex core the adapter depends on, not from the
adapter bundle (which names one model, `gpt-5.6-luna`, for its title
generator). 1.8.0 depends on `@openai/codex ^0.152.0`; caret on a `0.x` pins
the minor, so that resolves no higher than 0.152.1 and can never reach a 0.153
core. 1.10.0 depends on `^0.153.3`, which has the model.

Probed both over ACP stdio. 1.10.0 answers `initialize` and serves
`session/new` with the model config option listing:

```
gpt-6-astra, gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna, gpt-5.5, gpt-5.4,
gpt-5.4-mini, gpt-5.3-codex-spark
```

1.8.0 does not list `gpt-6-astra` at all.

Nothing else in 1.9.0 or 1.10.0 moves a surface codeg reads:
`sessionCapabilities` is byte-identical to 1.8.0's, and both of the new wire
surfaces stay inert (1.9.0's `_auth/status_update` lands on the
unrecognized-ext-notification arm; 1.10.0's async-task variants are gated on an
AIR advertisement codeg does not send to Codex). Per-version analysis is in the
registry comment as usual, along with the two comments 1.10.0 invalidated.

### DIFF
--- a/src-tauri/src/acp/connection.rs
+++ b/src-tauri/src/acp/connection.rs
@@ -3832,20 +3832,31 @@ fn build_client_capabilities(
         meta.insert("subagent-transcript".to_string(), serde_json::Value::Bool(true));
     }
     // claude-agent-acp 0.73.0 added "asyncTasks", and it is advertised — to
-    // claude ONLY, because codex-acp 1.8.0 does not implement the channel
-    // (its bundle contains no `asyncTasks` string at all). It publishes the
-    // lifecycle of Claude's NON-AGENT background work (background shells,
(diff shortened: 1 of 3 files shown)


Adam-Dalloul:
Closing: main already pins codex-acp 1.10.0 with your own analysis, so nothing left here.

PR #664, opened 2026-09-06
fix(acp): make forced command color opt-in instead of always on

chadnorvell:
## Problem

Every agent launch injected `CLICOLOR_FORCE=1` into the agent's process environment (`acp/connection.rs`, `DEFAULT_COMMAND_COLOR_ENV`). Because the agent runs its own bash tool in-process, every command it spawns — and everything those commands pipe into — inherited it.

`CLICOLOR_FORCE` is, by ecosystem convention, the one color variable `NO_COLOR` cannot override. `gh` (via go-gh) computes `colorEnabled = forced || (!disabled && isTTY)`, so forcing wins outright. Measured with gh 2.98.0, stdout on a pipe:

| env | `gh repo view --json name` |
|---|---|
| `CLICOLOR_FORCE=1 NO_COLOR=1` | `^[[1;37m{^[[m …` — ANSI *inside* the JSON |
| `CLICOLOR_FORCE=0` | `{"name":"codeg"}` |
| unset | `{"name":"codeg"}` |

So an agent running `gh … --json … | jq` got unparseable output, with no way to opt out from inside the session.

The quieter cost: those escapes are captured into the **agent's own context**. Every command an agent runs spent tokens on escape sequences and risked the model misreading output it needed to parse — charged on every turn, whether or not anyone looked at the colored card.

## Why not just scope the injection more narrowly

codeg can't. The colored output in the transcript's `<Terminal>` card comes from the agent's own command tools (`buildCommandTerminalOutput` in `content-parts-renderer.tsx`), and those processes are spawned *inside the agent*, not by codeg. The output codeg renders and the output the agent pipes into `jq` are the same process tree; the agent's environment is the only lever. Scoping the force to the one path codeg *does* spawn (ACP `terminal/create`) would relocate the bug, not fix it.

## Change

The injection is now **opt-in, defaulted off** — General Settings → "Colorize command output".

- `SystemTerminalSettings.colorize_command_output` (new field on the existing row; `#[serde(default)]` means pre-existing rows load as `false`, no migration).
- `merge_agent_env` consults a process global mirroring the setting, applied at startup and on every save. It's a global rather than a threaded handle because `merge_agent_env` is sync with a dozen callers and already reads two other ambient sources the same way (`proxy::current_proxy_env_vars`, `prepend_officecli_path`).
- `merge_agent_env_with_color` is split out pure so the behavior is testable without writing the global.
- Precedence is unchanged: a per-agent `CLICOLOR_FORCE` in the agent's env row still outranks the default either way.
- `apply_persisted_terminal_shell_config` → `apply_persisted_terminal_settings`, since it now applies both live values from one load.

Turning the switch on surfaces what it costs, inline, where it was turned on.

## Follow-up (not in this PR)

The ACP `terminal/create` path gets no color at all, before or after this change. Giving that runtime a real PTY (codeg already depends on `portable-pty` for the interactive terminal panel) would colorize it *correctly*: a bare `gh pr list` sees a TTY and colorizes, while `gh --json … | jq` has a pipe on stdout and doesn't. That's the version of this feature that costs nothing, but it only covers the path codeg spawns.

## Test plan

- `cargo test --no-default-features --bin codeg-server --lib` — new coverage for the default-off launch env, the opt-in, per-agent override precedence, save/restart round-trip, and legacy-row parsing.
- `pnpm test`, `pnpm eslint`, `tsc --noEmit`
- New frontend test pins the switch defaulting to off and the save carrying `default_shell` back unchanged (both settings share one stored row).

Three unrelated tests fail on this host both with and without the change (`/bin/cat` and `/bin/echo` don't exist on NixOS; a stray `/tmp/.git` breaks the "no `.git` above" fallback case). Verified identical on `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


xintaofei:
codeg work task `65` is done — https://github.com/xintaofei/codeg/pull/664 (15 files, +196/-30).

PR #667, opened 2026-09-06
fix(tabs): reopen a closed tab at its original position

Adam-Dalloul:
Ctrl/Cmd+Shift+T brought a closed tab back at the end of the strip instead of where it had been.

The closed-tab entry now records the slot the tab was closed from, and reopening splices it back in there, clamped to the current strip length, the way a browser restores a tab. Close-others and close-all record each tab against the strip as it would be once the earlier ones are gone, so repeated reopens walk the strip back into its original order. Conversation, draft and file tabs all go through it; activation, the dedupe of an already-open conversation, the preview slot rule and the stack cap are unchanged.

Follow-up to #502.

### DIFF
--- a/src/components/layout/workspace-chrome-controller-source.test.ts
+++ b/src/components/layout/workspace-chrome-controller-source.test.ts
@@ -66,6 +66,12 @@ describe("tab close/navigation shortcuts live in the always-mounted controller",
     )
   })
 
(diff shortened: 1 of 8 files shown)


xintaofei:
codeg work task `67` is done — https://github.com/xintaofei/codeg/pull/667 (3 files, +132/-7).

PR #670, opened 2026-09-06
fix(cline): close a stripped block at its own closing tag

MaxFreedomPollard:
## Summary

- search each block's closing tag in the slice after its own opening tag, in `strip_environment_details` and `extract_feedback` (`src-tauri/src/parsers/cline.rs`)
- end a `# task_progress RECOMMENDED` block at whichever section boundary comes first, instead of preferring `\n<` wherever it is
- add the file's first unit tests: the four broken shapes plus the well-formed, nested and unclosed ones

## Why

Cline wraps what the user typed in `<task>` and appends `<environment_details>` after it, so the user's own words share one string with the markers this parser strips. Both loops in `strip_environment_details` found the opening tag and then searched for the closing tag from the start of the whole message, so a message that quotes one of these tags puts a closing tag ahead of the block it appears to close.

For `<environment_details>` that is a hang. With `end < start`, `result[..start] + result[end..]` splices the opening tag back into the result and returns a string longer than the one it was given, so `while let Some(start) = result.find(...)` never terminates: `get_conversation` for that task spins and grows the string until the process runs out of memory. A message as ordinary as `Why do I see </environment_details> in my logs?` does it, and there is no way back to that conversation afterwards.

For `<task>` the same slip reads `result[tag_end..close]` with `close < tag_end` and panics on the reversed range: `begin > end (23 > 0) when slicing`. `extract_feedback` was already guarded by `end > inner_start`, so it returned `None` and dropped the feedback rather than crashing.

Searching from after the opening tag also makes both loops strictly shrink the string, so neither can spin on any input. Well-formed messages come out exactly as before, nested `<task>` wrappers still collapse to their innermost content, and an unclosed opener still truncates the way it did.

The `# task_progress RECOMMENDED` boundary is a separate slip in the same function. `rest.find("\n<").or_else(|| rest.find("\n#"))` takes the next tag whenever there is one and only falls back to a heading, so a `#` heading sitting between the block and the next tool tag was deleted along with the block. The comment says "next section", and the next section is whichever of the two comes first.

## Validation

- `cargo test --features test-utils` in `src-tauri`: green, 3531 lib tests plus every integration target, `parsers_snapshot` included with no snapshot change
- `cargo test --features test-utils --lib parsers::cline::`: 7 passed
- the same 7 tests compiled against unmodified `main`: `a_quoted_closing_tag_does_not_hang_the_environment_strip` runs until it is killed, `a_leading_closing_task_tag_does_not_panic` panics with `begin > end (23 > 0)`, `the_task_progress_block_ends_at_the_first_boundary` and `feedback_is_read_from_its_own_closing_tag` fail on the value, and the three well-formed cases pass on both sides
- `pnpm eslint .`: no error; `pnpm test`: 424 files, 6105 tests passed
- `cargo clippy --all-targets --features test-utils -- -D warnings`: no finding in `parsers/cline.rs`

### DIFF
--- a/src-tauri/src/parsers/cline.rs
+++ b/src-tauri/src/parsers/cline.rs
@@ -639,6 +639,7 @@ fn strip_environment_details(text: &str) -> String {
 
     // Remove <environment_details>...</environment_details>
     while let Some(start) = result.find(ENV_OPEN) {
+        let before = result.len();
         let inner_start = start + ENV_OPEN.len();
         if let Some(close) = result[inner_start..].find(ENV_CLOSE) {
             let end = inner_start + close + ENV_CLOSE.len();
@@ -647,6 +648,16 @@ fn strip_environment_details(text: &str) -> String {
             // Unclosed tag — remove from start to end
             result = result[..start].to_string();
         }
+        // Searching the closing tag from `inner_start` is what keeps every
+        // pass strictly shorter, and a pass that does not shrink is a pass
+        // this loop repeats forever. Asserted rather than left implied
+        // because the regression it guards grew the string exponentially: a
+        // test that trips it again would hang the run and exhaust memory
+        // instead of failing, and this fails it on the first pass.
+        debug_assert!(
+            result.len() < before,
+            "environment strip must shrink the message on every pass"
+        );
     }
 
     // Remove <task>...</task> wrappers, keeping inner content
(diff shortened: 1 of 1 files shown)


xintaofei:
codeg work task `66` is done — https://github.com/xintaofei/codeg/pull/670 (1 file, +66/-0).

PR #672, opened 2026-09-06
feat(composer): act on the token under a right click

Adam-Dalloul:
Right-clicking an email address, a link or a path in the composer opens the menu with Cut and Copy greyed out: both are gated on a non-empty selection, and a right click only moves the caret. The only way to copy one today is to highlight it by hand first.

A right click over the editing surface now selects the token under the pointer before the menu opens, so Cut and Copy apply to it. The token is the whitespace-delimited run with its wrapping punctuation peeled off, classified as an address, a link, a path or a plain word by a small pure helper. A right click inside a selection you already made leaves that selection alone, the way a native text field behaves.

Where the token has somewhere to go, one row is added above the editing items: open a link, write to an address, open a local file. It goes through the same opener the transcript uses for its links, so the protocol allow list, the desktop/web/remote routing and the failure toasts are the ones already in place. A plain word, or a path that opener cannot resolve on its own, gets no row.

On which menu: the composer already replaces the native one with the radix menu wherever the async clipboard read exists, and that menu carries Paste as plain text and Quick messages the native one cannot offer, so this keeps it rather than handing the box back to the browser. The selection step deliberately sits outside the radix trigger, so in a non-secure web context, where the custom menu stays disabled and the browser own menu appears, that one now opens over a selected token too.

Left out: the rendered transcript. A link in a message is caught by the conversation panel own context menu and plain user text is not linkified at all, so covering it needs a second implementation over DOM ranges plus a menu-collision rule. The transcript already has the selection bubble for copy and quote.

Tests cover the helper (address, link with and without a scheme, Windows and POSIX paths, token boundaries, empty input, whitespace), the editor glue (caret fallback, selection preserved, no crossing a node), and the menu itself.


xintaofei:
codeg work task `68` is done — https://github.com/xintaofei/codeg/pull/672 (6 files, +139/-5).

PR #681, opened 2026-09-07
fix(tabs): keep tiled panes when a preview tab opens

Adam-Dalloul:
Tile Display shows every tab in a group as its own pane at the same time, and the shortcut tip in Settings describes it as the way to see several sessions at once. Opening a conversation from the sidebar with a single click still takes the group's preview slot, so it replaces one of those panes in place and the session that was sitting there disappears from the tiling.

In a tiled group a preview open now takes a slot of its own instead, the same way a pinned open does.

Untiled groups are untouched: a single click still replaces the group's one preview, a double click still pins, and a conversation that is already open is still focused rather than opened a second time. The new tab stays unpinned either way, so it still reads as a preview and a double click still promotes it.

### DIFF
--- a/src/contexts/tab-context.test.tsx
+++ b/src/contexts/tab-context.test.tsx
@@ -2786,4 +2786,137 @@ describe("TabProvider tab groups", () => {
       expect(isReparentUnmount(store(), "conv-1-codex-1", home)).toBe(false)
     })
   })
+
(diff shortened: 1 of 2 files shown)


xintaofei:
codeg work task `180` is done — https://github.com/xintaofei/codeg/pull/681 (1 file, +3/-2).

PR #683, opened 2026-09-07
feat(tabs): a setting that keeps an opened conversation in its own tab

Adam-Dalloul:
Opening a conversation from the sidebar replaces the last one you opened, in
place. That is fine for a quick look, but a conversation only becomes a real
tab once you send a message in it, so the ones you open just to read (usually
the ones that have already finished) are the ones that get thrown away, and
they disappear without you asking for anything.

This adds "Keep opened conversations" to Appearance settings, under a new
Conversation tabs section next to the mode selection area. It is off by
default, so nothing changes unless you turn it on. With it on, a click opens a
tab that stays until you close it.

A kept tab is marked pinned rather than just spared. The strip sets a preview
in oblique type, and a tab nothing will replace should not be wearing that
mark, nor become replaceable again if the setting is later turned back off.
Double click still pins, an already-open conversation still focuses instead of
opening twice, and sending a message still pins.

The setting is read in `openTab` before the preview slot is considered, so it
sits beside the tiled-group check in #681 rather than on top of it. The two
branches merge with no conflict and both test blocks pass together.

Strings are in all ten locales.


xintaofei:
codeg work task `181` is done — https://github.com/xintaofei/codeg/pull/683 (75 files, +11345/-923).

PR #691, opened 2026-09-08
fix(chat): render local Markdown images inline

Adam-Dalloul:
Local screenshot paths in assistant Markdown currently render as `Image blocked` placeholders. This displays workspace images inline and opens them in the existing image viewer with copy and download actions.

Relative paths, absolute Windows/POSIX paths, file URIs, and reference images use the existing workspace-confined reader. Each transcript owns its image root, including inactive tabs and embedded sessions. Reads are limited to 8 MiB and four concurrent requests; Markdown's URL sanitization stays in place.

Frontend CI passes: 6,283 tests, lint, and the static export build. Browser checks cover inline display, the image viewer, light/dark themes, and a 360px layout.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.1.0",
     "@testing-library/user-event": "^14.5.2",
+    "@types/mdast": "^4.0.4",
     "@types/node": "25.2.2",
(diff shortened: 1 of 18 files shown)


xintaofei:
codeg work task `169` is done — https://github.com/xintaofei/codeg/pull/691 (9 files, +240/-8).

PR #701, opened 2026-09-09
fix(acp): reap child when monitor future is never polled

asteroida123:
## Summary

- construct `ChildGuard` before returning the child-monitor future
- ensure dropping that future before its first poll still kills and reaps the child
- add a Unix regression test that verifies `on_exit` fires exactly once

## Why

`connect_to` races the protocol driver with the child monitor. If the protocol side completed before the monitor future was first polled, the previous `async fn` body had not constructed `ChildGuard` yet. Dropping the unpolled future therefore dropped the raw child without the detached reap path or exit callback.

Extracted from #693 so the process-lifecycle fix can land independently.

## Verification

- `cargo test --manifest-path src-tauri/vendor/sacp-tokio/Cargo.toml --lib` (20 passed)
- `cargo test --manifest-path src-tauri/vendor/sacp-tokio/Cargo.toml --lib --locked`
- `git diff --check`

### DIFF
--- a/src-tauri/vendor/sacp-tokio/src/acp_agent.rs
+++ b/src-tauri/vendor/sacp-tokio/src/acp_agent.rs
@@ -377,8 +377,9 @@ impl AcpAgent {
     /// Register a callback invoked once with the OS process id (pid) of the
     /// spawned agent process, right after it launches.
     ///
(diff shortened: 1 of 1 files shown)


xintaofei:
codeg work task `172` is done — https://github.com/xintaofei/codeg/pull/701 (1 file, +25/-4).

PR #702, opened 2026-09-10
fix(codex): restore semantic MCP cards from code-mode history

asteroida123:
## Summary

- reconstruct code-mode MCP calls from persisted semantic McpToolCall records
- preserve real MCP call IDs, arguments, results, and error state when static script evaluation cannot resolve variables
- recover deferred scripts on their terminal wait only when one logical script is unambiguous
- keep native collaboration capsules separate from semantic MCP delegation cards

## Why

Codex code mode can execute multiple MCP calls behind one outer exec wrapper and print their results as one JSON chunk. The static script parser cannot always evaluate variable-based arguments, so history reload falls back to a generic script card and loses the real MCP call identity.

This uses the app-server semantic completion records as the authoritative source while retaining conservative fallbacks whenever correlation is ambiguous.

Extracted from #693 so the parser fix can land independently.

## Verification

- cargo test --no-default-features --lib --features test-utils parsers::codex (220 passed, 1 ignored)
- targeted semantic split, native/semantic identity, deferred completion, and late-binding regressions
- cargo check --no-default-features --lib
- git diff --check

### DIFF
--- a/src-tauri/src/models/message.rs
+++ b/src-tauri/src/models/message.rs
@@ -129,7 +129,7 @@ pub enum ContentBlock {
         /// call already completed — so copying it onto recovered inner calls
         /// would manufacture a permanent spinner. The one codex card that does
         /// carry a status is an MCP call rebuilt from its OWN semantic
-        /// `item_completed` record, which states a per-call terminal outcome
+        /// `item_completed` record, which yields a per-call terminal outcome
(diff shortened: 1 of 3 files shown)


xintaofei:
codeg work task `173` is done — https://github.com/xintaofei/codeg/pull/702 (2 files, +717/-32).

PR #707, opened 2026-09-10
fix(import): group an imported worktree session under its repo

### LINKED ISSUE #552: Bug: Imported Codex worktree sessions are registered as top-level folders instead of under repo root
When importing Codex sessions whose `cwd` is a Git worktree of an already-open repo, CodeG sometimes registers the worktree folder as a top-level workspace folder instead of grouping it under the repo root.

Example local state:

- Root repo: `C:\_Data\Projects\WowApps`
- Worktree grouped correctly:
  - `C:\_Data\Projects\worktrees\churn-report-edit-entries`
- Worktrees shown incorrectly as top-level:
  - `C:\_Data\Projects\worktrees\admin-danger-zone-edit-area-on-vendor-edit`
  - `C:\_Data\Projects\worktrees\pricing-investigation`

DB evidence:

- `folder.id = 32`, path `...\churn-report-edit-entries`, `parent_id = 1`
- `folder.id = 33`, path `...\admin-danger-zone-edit-area-on-vendor-edit`, `parent_id = NULL`
- `folder.id = 34`, path `...\pricing-investigation`, `parent_id = NULL`

All three are valid Git worktrees from `C:\_Data\Projects\WowApps` according to `git -C C:\_Data\Projects\WowApps worktree list --porcelain`.

Likely cause:

The normal worktree registration path uses `open_worktree_folder_core`, which writes the parent repo id via `folder_service::add_folder_with_parent(...)`.

But the session import path in `src-tauri/src/commands/conversations.rs` calls plain `folder_service::add_folder(...)` for imported session `cwd` paths. That creates/reopens a normal folder and leaves `parent_id = NULL`. It does not resolve whether the imported `cwd` is a Git worktree belonging to an existing root folder.

Expected behavior:

When importing a session whose `cwd` is a Git worktree of an already-open folder, CodeG should register/reconcile that folder as a worktree child of the root repo, same as `open_worktree_folder_core` does.

Relevant code:

- `src-tauri/src/commands/folders.rs`
  - `open_worktree_folder_core(...)`
- `src-tauri/src/db/service/folder_service.rs`
  - `add_folder(...)`
  - `add_folder_with_parent(...)`
- `src-tauri/src/commands/conversations.rs`
  - `import_selected_from_summaries(...)` currently calls `folder_service::add_folder(...)`

Adam-Dalloul:
Fixes #552.

Importing a session whose cwd is a linked git worktree registers the worktree as one more top-level folder, next to the repo it belongs to instead of under it. `import_selected_from_summaries` calls `folder_service::add_folder`, which is `ParentWrite::Preserve` and inserts `parent_id = NULL`, and `parent_id` is exactly what the sidebar reads to tell "worktree of this repo" from "another project". The branch-label backfill selects on `parent_id IS NOT NULL` as well, so such a folder never picks up its branch name either.

`git_repo::main_worktree_root` resolves the repo a worktree belongs to from git's own on-disk pointers rather than by spawning git: `<wt>/.git` is a file holding `gitdir:`, the private dir it names carries `commondir`, and for a non-bare repo that shared dir is `<main working tree>/.git`. A submodule's git dir has no `commondir` and a bare repo's common dir is not named `.git`, so neither resolves to a parent.

The importer uses `add_folder_with_parent` only when that root is a folder codeg already has open, and otherwise keeps `add_folder`. So a non-worktree cwd is unchanged, a worktree of a repo the user never opened stays top-level rather than inventing a workspace row, and a reopen still cannot demote a folder that a worktree open already parented.

Verified against a real `git worktree add` on Windows, including the sibling layout from the report where the worktree lives outside the repo directory.

### DIFF
--- a/src-tauri/src/commands/conversations.rs
+++ b/src-tauri/src/commands/conversations.rs
@@ -988,8 +988,8 @@ pub(crate) async fn import_selected_from_summaries(
                 // off — the sort above put every plain folder ahead of every
                 // worktree precisely so this lands in time. Only plain folders
                 // are recorded: a linked worktree is never anyone's main working
-                // tree. `add_folder` left an existing row's `parent_id` alone, so
-                // the flattening rule is the seed's, applied to the row as read.
+                // tree. The candidate is flattened by the seed's rule, off the
+                // parent this row actually ended up with.
                 if root_key.is_none() {
                     let root = persisted_parent_id.unwrap_or(folder_id);
                     worktree_parents.insert(norm_key, root);
--- a/src-tauri/src/commands/conversations.rs
+++ b/src-tauri/src/commands/conversations.rs
@@ -532,6 +532,30 @@ fn index_folder_rows(rows: &[ScanFolderRow]) -> HashMap<String, &ScanFolderRow> 
     index
 }
 
+/// The folder id an imported session's cwd should hang off, or `None` to leave
+/// the folder top-level.
+///
+/// `Some` only when the cwd is a linked git worktree AND its main working tree
+/// is a folder codeg already knows: grouping under a repo the user never opened
+/// would invent a workspace row, and a non-worktree cwd is top-level by
+/// definition.
+fn worktree_parent_folder_id(
+    path: &str,
+    folder_index: &HashMap<String, &ScanFolderRow>,
+) -> Option<i32> {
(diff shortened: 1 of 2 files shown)


xintaofei:
codeg work task `175` is done — https://github.com/xintaofei/codeg/pull/707 (1 file, +456/-48).

PR #714, opened 2026-09-10
fix(tasks): a requeued task starts over instead of resuming the run you cancelled

### LINKED ISSUE #649: 待办任务取消流转问题
当一个任务点击开始，然后点击取消，再点击重新排队，再点击开始，这个任务会转一会圈就失败了跑到已完成里。已经复现很多次。只能重新新创建一个任务。

Adam-Dalloul:
Fixes #649: start a task, cancel it, requeue it, start it again, and it spins for a while and then lands in 已完成 as `canceled`. The same task never runs again; only a brand new task works.

Cause: `requeue_canceled` moves the row back to `todo` and clears the plan, the failure reason and the pending merge, but keeps `conversation_id`. `launch_mode_for` reads exactly that column to choose between `Retry` and `Fresh`, so the next start is dispatched as a resume of the run the user just cancelled: it loads the killed agent session, and it reuses the conversation row whose status is still `cancelled`. The 30 second reconcile sweep reads that stale status as this generation's verdict and settles the task `canceled`, which is the 已完成 column. Nothing ever clears `conversation_id`, which is why a new task is the only way out.

The change: a requeue drops the conversation link. The worktree is still reused, and the old conversation still exists in its folder; only the session pointer goes, so the next start runs the task from the top. `retry` (failed to queued), which deliberately continues the same session, is untouched, and the test asserts that.

The first commit is the failing test on its own, so the CI run on it shows the defect: `a_requeue_drops_the_canceled_run_s_conversation` reads `Some(41)` where it expects `None`.

One consequence worth calling out: while a requeued task waits in `todo`, the "View session" footer action is hidden, because there is no bound conversation until the next run starts. That matches what the task now is, a fresh unit of work on the board, but it is a visible change.

### DIFF
--- a/src-tauri/src/db/service/work_task_service.rs
+++ b/src-tauri/src/db/service/work_task_service.rs
@@ -3226,6 +3226,80 @@ mod tests {
         assert_eq!(actions, vec!["schedule", "unschedule"]);
     }
 
+    /// #649: start, cancel, requeue, start again used to land the task in
+    /// `canceled`. The requeue kept the conversation of the run the user had
+    /// just stopped, and `launch_mode_for` reads exactly that column to pick
+    /// `Retry` over `Fresh`, so the second start resumed the killed session
+    /// instead of running the task, on a conversation row still recorded as
+    /// `cancelled`, which the reconcile sweep then read as this generation's
+    /// verdict. A requeue puts the task back on the board, so it keeps the
+    /// worktree and drops the session.
+    #[tokio::test]
+    async fn a_requeue_drops_the_canceled_run_s_conversation() {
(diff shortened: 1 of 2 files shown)


Adam-Dalloul:
The CI run on the first commit (test only) is the reproduction: `a_requeue_drops_the_canceled_run_s_conversation` failed with "a requeued task must start over, not resume the canceled session", 3574 passed / 1 failed. The second commit is the fix.

xintaofei:
codeg work task `202` is done — https://github.com/xintaofei/codeg/pull/714 (2 files, +387/-60).

PR #719, opened 2026-09-11
fix(tests): make the pi sessionDir test use a platform-absolute path

Adam-Dalloul:
`parsers::pi::tests::settings_json_session_dir_is_honored` fails on `Rust server on windows-latest`, and it fails on main too (run 34542650449 at a6918223), so every open PR inherits one red job.

The test hard-codes `/srv/pi-sessions`. `session_dir_from_settings` only honors an absolute `sessionDir`, and `Path::is_absolute` wants a drive or UNC prefix on Windows, so that literal is not absolute there. The gate correctly refuses it, the resolver falls back to `<agent_dir>/sessions`, and the assertion fails. The `~`-rooted case right below it has the same problem for the same reason, it just never got to run.

The product code looks right to me, so this only changes the test: both paths now get a `C:` prefix when the test runs on Windows. Verified by extracting the resolver and running it here with a windows-gnu toolchain: the current form fails both cases and the new form passes, alongside the env-var and relative-path cases which were never platform dependent.

### DIFF
--- a/src-tauri/src/parsers/pi.rs
+++ b/src-tauri/src/parsers/pi.rs
@@ -2311,25 +2311,39 @@ mod tests {
     /// pi-acp resolves `<agentDir>/settings.json` → `sessionDir` before falling
     /// back to `<agentDir>/sessions`, so a user who sets it kept a pi that
     /// resumed fine and a codeg history list that was permanently empty.
+    /// An absolute path on the platform the test is RUNNING on.
+    ///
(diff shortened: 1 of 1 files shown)


Adam-Dalloul:
Closing: c80b2161 fixes this, and more thoroughly. Building the paths through a helper and serializing the JSON also covers the backslash-escape case my version would still have tripped on.

PR #725, opened 2026-09-11
feat(desktop): register codeg:// URL scheme to open a conversation

changexbc:
## Why

The desktop app never registered a system URL scheme (`CFBundleURLTypes` / protocol handler). External tools can launch Codeg (`open -b app.codeg`) but cannot jump to a **specific** conversation tab.

In-app markdown already uses `codeg://session/<id>` as a mention badge. This PR makes that form a real OS deep link.

## What

- Register the `codeg` custom scheme via `tauri-plugin-deep-link` (macOS / Windows / Linux).
- Resolve:
  - `codeg://session/<id>` — Codeg numeric PK **or** the agent's `external_id` (Grok UUID, Codex thread id, …)
  - `codeg://workspace?conversationId=<id>` — same lookup; `folderId` / `agent` optional
  - `codeg://open` — just show the workspace
- Ignore in-app mention badges (`codeg://agent/…`, `codeg://commit/…`, `codeg://embedded/…`).
- **Warm start:** reuse `workspace://focus-conversation` (same path as the pet panel).
- **Cold start:** bake the lookup into `/workspace?folderId=&conversationId=&agent=` so the existing `DeepLinkBootstrap` opens the tab after hydration (an event emitted before the webview subscribes would be dropped).
- Second-instance argv on Windows/Linux is forwarded through the single-instance plugin.

Docs: `docs/url-scheme.md`.

## Examples

```bash
open "codeg://session/214"
open "codeg://session/<grok-or-codex-session-id>"
```

```js
window.location.href = "codeg://session/214"
```

## Tests

- Parser / argv / lookup unit tests in `src-tauri/src/deep_link.rs` (numeric id, external id, path traversal, workspace query mismatch).
- `cargo test --features test-utils --lib deep_link`
- `cargo clippy --lib --features test-utils -- -D warnings`

Not available in `codeg-server` / browser-only mode — that path already has the `/workspace?…` query string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ Step away from your desk, not your work. The native iOS and Android clients conn
 - **[To-dos](https://docs.codeg.app/guide/tasks)** — write down what needs doing and agents work through the queue, each task in its own worktree, landing on your branch only after you've reviewed it
 - **[Custom Agents](https://docs.codeg.app/guide/custom-agents)** — register any other ACP-compatible agent from the public registry or its distribution JSON; Codeg installs it, records its history, and treats it like a built-in
 - **[The Workspace](https://docs.codeg.app/guide/workspace)** — the full engineering loop next to the agent: file tree, editor and diff, git changes, commit, an embedded terminal, and [several folders linked into one workspace](https://docs.codeg.app/guide/workspace#work-across-several-folders)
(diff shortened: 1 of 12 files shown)


xintaofei:
codeg work task `96` is done — https://github.com/xintaofei/codeg/pull/725 (12 files, +905/-128).

PR #727, opened 2026-09-11
fix(composer): badge a slash word only when it is a real command

Adam-Dalloul:
Typing `/` and a word made it a slash command: text the composer fills itself with (a paste, a restored draft, a quick message, a reopened queued message, a to-do brief) turned any `/word` into a command badge on shape alone, whether or not the agent had such a command and whether or not anyone picked it from the menu.

A badge now needs the token to be one the agent advertises right now, matched exactly, drawn from the same lists behind the `/` and `$` menus. `/rev` is not `/review`; `/Review` is not `/review`, since the CLI reads the name it advertised. With no list yet (connection still coming up, or no agent behind the box) nothing is badged, which is the safe direction.

What a badge changes about the message: its text is rebuilt on send from the badge's id rather than carried through, so an id the serializer has to escape (`/foo_`) goes out as `/foo\_`. Plain text goes out exactly as written, and stays editable in place instead of becoming one atomic node whose `/` is hidden.

Unchanged: picking from the menu with Enter, Tab or a click, including the trailing space; badges already in the document, which a list arriving mid-compose never rewrites; `file:`/`codeg:` reference links, which hydrate regardless because they were inserted deliberately; and the sent-message transcript, where the text is already fixed.

### DIFF
--- a/src/components/automations/automation-editor.tsx
+++ b/src/components/automations/automation-editor.tsx
@@ -379,6 +379,7 @@ export function AutomationEditor({
           mentionUiLabels={mentionUiLabels}
           tabLabels={referenceGroupLabels}
           mentionAnchorRef={composerBoxRef}
+          knownInvocations={invocations.knownInvocations}
           onChange={(text) => {
             setPrompt(text)
             invocations.detect()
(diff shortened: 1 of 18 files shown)


xintaofei:
codeg work task `94` is done — https://github.com/xintaofei/codeg/pull/727 (2 files, +125/-4).

PR #756, opened 2026-09-16
feat(chat): recall sent prompts with ArrowUp/ArrowDown

### LINKED ISSUE #744: [Feature Request] Use ArrowUp/ArrowDown to recall sent messages in the composer
## Motivation

In Codeg 0.30.7, when the message composer is focused, pressing ArrowUp does not put the latest sent prompt back into the input. This makes iterative prompts slower.

## Expected behavior

- ArrowUp recalls the latest sent message; repeated presses move to older messages.
- ArrowDown moves forward and restores the draft at the newest entry.
- Multiline messages keep normal caret movement until the first/last line boundary.
- CJK IME 候选导航和 / 命令补全保持正常。

Related: #7 (closed, unmerged PR).

## Environment

- Codeg 0.30.7
- Windows

dawNotPoi:
The chat composer can now step back through the current session's sent prompts with ArrowUp/ArrowDown, terminal-style.

Interaction (agreed on #744):

- A bare ArrowUp with the caret on the first line steps to the previous sent prompt; ArrowDown with the caret on the last line steps to the next one. Anywhere else the arrows keep moving the caret, so a multi-line prompt stays editable line by line.
- A recall lands on the edge you travelled FROM — the top for ArrowUp, the bottom for ArrowDown — so pressing the same key again keeps stepping without walking the caret first.
- ArrowDown past the newest prompt restores the draft that was in the composer before the first recall.
- Editing the box ends the recall; the next ArrowUp starts again from the newest prompt.
- The history is the current session's own user turns, so a new session has none and a resumed one gets its real prompts back; nothing is persisted separately.
- Only the chat composer: the task and automation editors pass no handler and keep the editor's default Arrow behaviour. IME composition and the `/`·`$` menus still own the keys first.

Implementation: the pure transition rules live in `src/lib/composer-history.ts`; `RichComposer` only reports edge-hitting Arrow keys (keeping the ProseMirror boundary logic out of the host); `MessageInput` owns the recall state and the draft stash. The history source is a getter, so the runtime store's per-token updates never recompute it or re-render the composer.

Closes #744.

Validation:
- New unit tests for the transitions and the turn extraction (9), plus `RichComposer` edge/IME/menu/modifier routing (8) and `MessageInput` integration on a real Tiptap editor (multi-line stepping, draft restore, edit-ends-recall, queue-edit guard).
- Full frontend suite: 444 files / 6,458 tests passed; TypeScript, ESLint and the static-export build passed.
- Manually exercised on `codeg-server` with an imported session holding multi-line user prompts.


dawNotPoi:
@xintaofei — implementing the interaction we agreed on in #744. A short summary so it can be checked against your intent:

1. Arrow keys only touch history **at the edge**. `↑` with the caret on the first line steps to the previous sent prompt, `↓` with the caret on the last line steps to the next one; anywhere else they keep moving the caret, so a multi-line prompt stays editable line by line.
2. A recall lands on the edge you travelled **from** — the top for `↑`, the bottom for `↓` — so the same key keeps stepping without walking the caret first. One consequence: turning around means walking the caret to the other edge first, which I read as your "scroll back down".
3. `↓` past the newest restores the draft that was in the box before the first recall. Editing the box ends the recall; the next `↑` starts again from the newest prompt.
4. The history is the current session's own user turns — nothing is persisted separately, so a new session has none and a resumed one gets its real prompts back (your point 1). Only the chat composer: the task and automation editors pass no handler and keep the editor's default Arrow behaviour. IME composition and the `/`·`$` menus still own the keys first.

Related but distinct: #476 restores one message by an Edit action and rewrites the transcript from there; this only fills the composer and leaves the transcript untouched.

Could you take a look when you have a moment? Happy to change the edge/landing rules if you would rather have it behave differently.

xintaofei:
codeg work task `107` is done — https://github.com/xintaofei/codeg/pull/756 (4 files, +129/-18).

PR #767, opened 2026-09-17
fix(macos): drain native fullscreen before close behavior

### LINKED ISSUE #507: macos系统下全屏点击关闭没有正确关闭
Image
全屏状态下点击关闭会留着黑色的空白和工具栏

Frank-zhu0404:
On macOS, clicking the main window's close button while the workspace is in native fullscreen (or still animating out of it) intercepts `CloseRequested` and applies hide / exit / ask without leaving the fullscreen Space. That leaves a black blank plus leftover toolbar chrome instead of dismissing the window.

The close path now drains native fullscreen first: `set_fullscreen(false)`, wait until tao's flag drops *and* AppKit has finished tearing down the Space (tao reports `is_fullscreen() == false` at `windowWillExitFullScreen`, which is the start of the animation), then apply the configured close behavior (`ask` / minimize-to-tray / `exit`).

Linux and Windows are unchanged — fullscreen there is not a separate Space, so close does not wait.

## Testing

- New unit tests for the occupancy / drain decision helper in `commands/windows.rs` (live fullscreen, transitioning after tao drops the flag, windowed, non-macOS never waits).
- Linux: `cargo test --features test-utils --lib commands::windows::` — 19 passed.
- Linux: `cargo clippy --all-targets --features test-utils -- -D warnings` passed.
- Linux: `cargo check --no-default-features --lib --bin codeg-server` passed.
- Native macOS GUI close-while-fullscreen was not exercised here.

Closes #507


xintaofei:
codeg work task `200` is done — https://github.com/xintaofei/codeg/pull/767 (3 files, +313/-291).

PR #768, opened 2026-09-18
feat(zcode): integrate ZCode as a built-in agent

asteroida123:
## What

Integrates **ZCode** as the sixteenth built-in agent, driven through the pinned community adapter [`zcode-codeg-adapter@0.1.3`](https://www.npmjs.com/package/zcode-codeg-adapter) (stdio ACP; the ZCode CLI itself ships no ACP server). This follows the DeepSeek Harness pattern — built-in enum entry + npm-distributed adapter — with one architectural difference called out below.

ZCode's own backend keeps its history in a private store, which codeg deliberately does not parse. **ZCode is therefore the first built-in whose history is codeg's own ACP recording** (`parsers::acp_native`, the same transcript path custom agents use) — `transcript_dir_for` now covers `Custom(_) | ZCode`. Every other built-in keeps its dedicated native-store parser.

## Notable decisions

- **Launch contract**: the adapter requires `ZCODE_CODEG_ENTRY` (absolute path to the ZCode app's `.cjs` entry) and exits before the handshake without it. codeg validates an explicit Agent Settings → ZCode env override, else probes the standard install locations (`/Applications/ZCode.app/...`, `~/Applications/...`); failure routes to an install-prompt error instead of an opaque mid-handshake exit.
- **Identity compatibility**: `zcode-codeg` stays a valid custom-agent id, so conversations persisted as `custom:zcode-codeg` (the pre-builtin registration) keep resolving; no forced migration.
- **MCP**: wire-only delivery (deliberately off the forward skip list — the codeg-mcp companion must reach it); no codeg-side native MCP store, so ZCode stays out of `McpAppType` and the MCP settings page. Skills stay undeclared until there is ZCode-side evidence of a load path.
- **`default_enabled`**: the variant is listed explicitly — the surrounding `matches!` compiles non-exhaustively, and without the arm the agent is born disabled and never appears in the picker.
- **READMEs**: built-in count fifteen → sixteen and ZCode joins the supported-agent roster, all ten languages.

## Live acceptance (ZCode CLI 0.16.5, macOS arm64)

- Multi-turn chat with context retention; streaming and clean turn settlement
- Tool cards for Read / Write / Edit / Bash with plan-mode read-only enforcement
- Permission cards (Allow once / Always allow in this project / Deny) — the adapter maps ZCode's native option kinds into the closed ACP enum; off-enum kinds used to fail client-side deserialization and tear the connection down
- Writes verified on disk; cancellation settles with `stop_reason=cancelled` and the session stays usable
- Server restart replays history exactly once (no duplicates) and the conversation continues via `session/load`
- Model selector lists the full desktop-config catalog (16 models across 9 providers on the test machine) with working switching on the 0.16.5 wire shape (`session/setModel` with `options.reasoningLevel`)

## Checks

- `cargo test --features test-utils` — 3953 passed
- `cargo test --no-default-features --bin codeg-server --lib` — clean
- `cargo clippy` ×3 modes with `-D warnings` — clean
- `pnpm test` — 6609 passed; `pnpm build` (static export) — clean
- New unit tests: launch-entry preflight matrix, transcript-recording coverage, delegation enum counts, registry pin assertions

The adapter itself is MIT-licensed, stdio-only, and pinned exactly; its audit trail (backend contract probes, dual-generation model-switch protocol, permission-kind mapping) lives in the [adapter repo](https://github.com/asteroida123/zcode-codeg-adapter).


asteroida123:
zcode暴露出了[上传工作区/.git记录到云端]的问题，可能快要开源了，说不定近期有大变化，可以等等看。

Issue #769, opened 2026-09-18
Intermittent sticky dragging when clicking session tabs on macOS


turingcat:
### Description

On macOS, clicking a session tab in the top tab bar can intermittently leave it in a sticky dragging state, instead of simply selecting the session.

### Steps to reproduce

1. Open multiple sessions in the desktop app.
2. Click between the session tabs in the top bar.
3. Occasionally a tab appears to remain in a drag state after the click.

The exact triggering sequence and frequency have not been isolated.

### Expected behavior

A click selects the session; dragging should end when the pointer is released.

### Environment and investigation

- Codeg 0.30.10, macOS, Apple Silicon (arm64).
- The original dependency declaration is Motion ^12.34.0.
- After upgrading Motion to ^13.4.0 and rebuilding the desktop app, the reporter has not observed the problem during initial manual use.

This is preliminary manual evidence, not confirmation of the root cause or a guaranteed fix. A dependency-only PR follows for review, including consideration of the Motion 12-to-13 major-version change.

PR #770, opened 2026-09-18
fix: upgrade Motion to address sticky session tab dragging

turingcat:
### Problem and change

Related issue: #769.

Clicking a top session tab can intermittently leave the tab in a sticky dragging state. Upgrade Motion from ^12.34.0 to ^13.4.0 and update pnpm-lock.yaml. Initial manual testing of the rebuilt macOS arm64 app has not reproduced the issue.

This crosses a major version boundary. The root cause has not been conclusively isolated, and the manual result should not be taken as proof that all drag interactions are fixed. No application behavior code is changed.

### Validation

- Next.js static export completed with `next build --webpack`.
- Tauri release build for aarch64-apple-darwin completed; app and companion binaries were verified as arm64.
- Reporter is manually testing the rebuilt macOS app and has reported no problems so far.
- No additional automated regression run was performed for this submission, at the reporter's request.


xintaofei:
codeg work task `199` is done — https://github.com/xintaofei/codeg/pull/770 (1 file, +129/-0).

PR #771, opened 2026-09-18
Show cache hit rate in the composer popover; read Pi's context window from models.json

gongcavalry:
## What

Two small gaps in the context meter, both hit by anyone running a model id that is not in the built-in name table.

### 1. The composer popover never states the cache hit rate

`ComposerContextUsage` lists input / output / cache read / cache write / total, but never the ratio. Judging cache efficiency from the composer means doing the division by hand, or leaving the conversation to open the global Token Usage page — which answers a different question (spend over time) than the one the popover invites (this session, right now).

This adds one row under **Total**: `Cache Hit Rate  94.6%`, computed as

```
cacheRead / (input + cacheRead + cacheWrite)
```

`cacheWrite` is kept in the denominator on purpose — dropping it treats repeated writes as free and inflates the ratio.

### 2. Self-hosted model ids get no context meter at all

`infer_context_window_max_tokens()` (`src-tauri/src/parsers/mod.rs`) matches model names against a fixed table — Claude / Gemini / Gemma / Kimi / Grok / OpenAI — and returns `None` for everything else. Any model served by a self-hosted OpenAI-compatible endpoint therefore always lands on `None`, and the consequences are visible:

- the trigger shows a raw token total (`5.7M`) where a percentage belongs, because `hasContext` is false;
- `context_window_max_tokens` is stored as absent, so the history gauge never appears either.

The number isn't missing — the window is simply unknown.

Pi already records the real window per provider in `~/.pi/agent/models.json` (`providers.<name>.models[].contextWindow`), so the parser now reads that first and only then falls back to guessing by name. Any read failure — file absent, bad JSON, model not listed, window 0 — yields `None` and takes the existing path, so nothing regresses for models the table already covers.

`FACT_SCHEMA_VERSION` bumps `2` → `3`. Without it the fix would reach only conversations touched afterwards, and the dashboard would keep serving a mix of old rows (no window) and new ones. There is already a test asserting a schema bump reparses stamped-current conversations; it still passes.

## Tests

| Check | Result |
|---|---|
| `cargo test --lib` (full) | 3949 passed / 0 failed / 1 ignored |
| New cases in `parsers::pi::tests` | 3 — largest window wins across providers, unknown id falls through, junk input yields `None` instead of panicking |
| `pnpm test` (vitest) | 6608 passed across 453 files |
| `pnpm lint` (eslint) | clean |
| `cargo check --lib` | 0 errors, 0 warnings |
| `pnpm build` + `tauri build --bundles app` | both succeed |

The new logic is a pure function over the file contents (`pi_declared_context_window_from`), so the tests cover it without touching the disk.

## Notes

- No new settings field. `custom_context_window` exists today but lives on `GrokSettings`, so it is not reachable for Pi — reading what Pi already declares avoids inventing a second place to configure the same number.
- i18n: `cacheHit` added to all ten locale files.
- No screenshot attached — the change is one text row plus a percentage where a token count used to be. Happy to add one if useful.

---

**中文说明**（仓库 README 有中文版，一并附上）：这两处都是「跑自建 OpenAI 兼容端点的人一定会碰到」的问题。第 2 条尤其容易被当成显示选项没开：模型名不在内置表里时，弹窗只能退化成显示绝对 token 数，历史侧的上下文仪表也整段消失，而设置面板里那个「上下文窗口」输入框只挂在 Grok 上，Pi 侧没有入口。改成优先读 Pi 自己声明的 `contextWindow`，比按名字猜准，换模型也自适应。


xintaofei:
codeg work task `204` is done — https://github.com/xintaofei/codeg/pull/771 (8 files, +525/-101).

PR #778, opened 2026-09-19
fix(acp): update session transcript pointer after /clear

### LINKED ISSUE #766: Session transcript pointer is not updated after /clear, hiding everything that follows
**What happens**

After `/clear` in a Claude Code session, everything that follows disappears from the panel when the session is reopened. The live turn still renders normally, so the loss only shows up later — which is what makes it expensive.

**Why**

`/clear` makes Claude Code roll over to a **new transcript file** with a new uuid. The ACP session id does not change: the adapter's `clear-context-coordinator` deliberately "replace[s] Claude's private conversation without replacing the public ACP turn" (`restartSession(params, { publicSessionId, … })`), and it never surfaces the new transcript path.

codeg derives the path from the session id (`acp_transcript.rs::transcript_path_in` → `<root>/<agent_dir>/<session_id>.jsonl`), which the log confirms:

```
[bg-watch] armed connection=<c> session=<id> baseline=<n> file=…\projects\<slug>\<id>.jsonl
```

So after `/clear`, codeg keeps watching a file nobody writes to, while `conversation.external_id` still points at it. The ACP connection itself stays healthy, which is why the problem only bites on reload.

**Repro**

1. New Claude Code session, send a message
2. `/clear`, send another message
3. Close and reopen the session — only step 1 is there

The step-2 content is intact in a new `<slug>/<new-uuid>.jsonl` whose first records contain `<command-name>/clear</command-name>`, starting at the timestamp the old file stops. Re-pointing `conversation.external_id` at that file makes it visible again immediately, with no restart — so the reader is the only thing missing.

**Note**

Fork already does the right thing: `acp/manager.rs` sets `external_id` to the forked session id. `background_watch` already has `rearm` / `adopt_file`, and there is an `update_external_id` service. The pieces look present; `/clear` just has no path into them.

A codeg-only fix would have to detect the rollover on disk, since the adapter reports no new id — having the adapter report the current transcript path over ACP would be the cleaner half of it.

Observed on codeg 0.30.8, Claude Code ACP 0.77.0; source references from the 0.30.7 tag.

— 胡淇华

Frank-zhu0404:
After `/clear` in a Claude Code session, Claude rolls over to a **new transcript uuid file** while the ACP session id stays the same. codeg kept watching the abandoned `{session_id}.jsonl` and left `conversation.external_id` pointing at it, so post-clear turns disappeared on reopen (the live turn still rendered).

This change detects the `/clear` successor on disk, has the background watcher adopt that file (ACP session id unchanged), emits `transcript_rolled_over`, and re-points `conversation.external_id` so reopen/read paths follow the chain.

## Testing

- Unit tests: successor pick, ignore unrelated/stale siblings, reopen follows chain to post-clear turns, watcher adopts new file while keeping ACP id, lifecycle external_id repoint (+ second rollover).
- `cargo test --features test-utils --lib -- clear_rollover` → **5 passed**.

Fixes #766


xintaofei:
codeg work task `214` is done — https://github.com/xintaofei/codeg/pull/778 (3 files, +372/-52).

PR #793, opened 2026-09-21
fix(science): include Antigravity in install status

### LINKED ISSUE #718: 给Google Antigravity添加【科学研究】技能包报错
第一次启用的时候，.gemini下有正常链接，但是界面显示开启后秒变关闭。
再次勾选启用，提示

Image

仅科学研究可一定复现此问题

dawNotPoi:
## Summary

- include Antigravity in the scientific research pack's authoritative install-status inventory
- keep the backend snapshot aligned with the skill-capable agent column shown by the frontend
- add a regression test that queries real per-skill install statuses and requires an Antigravity row

## Root cause

The settings UI includes Antigravity because ACP reports it as skill-capable, but the science backend used an older hard-coded agent subset. The first enable created the link successfully, then the authoritative refresh omitted Antigravity and flipped the toggle back off; a second enable retried against the existing link.

## Verification

- `pnpm build` — production static export passed (32/32 pages)
- `git diff --check` passed
- added `commands::science::tests::install_statuses_include_antigravity`
- full Rust execution is delegated to GitHub CI to avoid recreating a 25 GiB local Windows build cache after the development drive reached capacity

Closes #718


xintaofei:
codeg work task `213` is done — https://github.com/xintaofei/codeg/pull/793 (1 file, +54/-11).

PR #795, opened 2026-09-21
fix(files): keep deleted and dismissed Office previews closed

Adam-Dalloul:
Removing an Office document or its temporary directory can open a new preview of the deleted file and show "workspace root does not exist". The auto-preview listener treats every `changed_paths` entry as a document to open, including removals. Closing a preview and switching folders also resets its deduplication state, allowing the tab to return.

Check directory metadata before opening, share sibling lookups, and retain the seen paths across folder changes. Pending checks are cancelled when the active workspace or preference changes. This uses the existing directory API without reading document contents.

Added six regression tests. All 138 tests across the workspace context, Office preview, workspace-state store and file-path handling pass, along with lint and formatting for the changed files.


xintaofei:
codeg work task `222` is done — https://github.com/xintaofei/codeg/pull/795 (2 files, +67/-1).

